PR #5, opened 2015-02-04
Packagist update and installers

derhasi:
I put together a PR for the crucial changes of my fork:

* Replacing drupal packagist with webflo's instance
* Fetching drupal core from packagist
* Updating custom installer paths
* Adding derhasi/preserve-paths for contribs

We should rework the README.md on the base of that, but I'd like to cover that in a separate PR, as we could separate the README in different separate .md files (drush make, FAQ) and only provide a quick overview in the main README.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,18 @@
+# OS specific temporary files
+.DS_Store*
+ehthumbs.db
+Thumbs.db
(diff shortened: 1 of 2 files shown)


dutchiexl:
Looking good

PR #7, opened 2015-02-11
Remove files on create-project that are only relevant for testing.

derhasi:


kasperg:
:+1:

jcnventura:
@webflo Shouldn't this post-create-project-cmd hook also remove the same files in the 8.x branch?

PR #18, opened 2015-04-18
Make Devel a dev requirement.

andrewholgate:
Devel is a development tool so would be best included as a dev requirement.


yched:
sounds sensible ?

webflo:
Yes i think it makes sense, but i had some problems with the core autoloader in the past if a module was installed but not available.

webflo:
#38 is newer than this PR but contains the discussion about this feature.

Closed in favour of #38 

@andrewholgate Please follow the other issue.

PR #24, opened 2015-05-19
Update Drush requirement to "drush/drush":  "8.*"

greg-1-anderson:
Drush now has a 7.x branch, and 'master' is labeled '8.x-dev'.  In composer.json, we therefore now need to require drush/drush 8.\* to get the version that is compatible with Drupal 8.


tstoeckler:
`composer install` failed before, and goes through with the PR. Thanks a lot!

PR #31, opened 2015-06-29
.idea should be ignored globally

zkanda:
Hey guys, .idea is better ignored globally rather than per repo specific..
https://help.github.com/articles/ignoring-files/#create-a-global-gitignore


derhasi:
I do not think so. You have to think about working in a team on one project. You cannot control the gitignore of every team member. So if you want to avoid conflicts you better put the ignores in every project.

PR #34, opened 2015-07-09
Keep `web/` up to date

jonhattan:
"A script to update `web` /
and the actual update of `web` /
by running the script."

---

The script is based on drush for several reasons:
- composer seems actually unable to download beta12. I've tried with:

```
composer require drupal/drupal:@beta
composer require drupal/drupal:8.0.0-beta12
```

It complains with `drupal/drupal 8.0.0-beta12 requires drupal/core ~8.0 -> no matching package found.` Perhaps I'm missing something obvious here.
- `drush dl` downloads a tgz, quite quicker than using composer to download `drupal/drupal` package and its dependencies
- `drush dl` downloads automatically the last version, no matter its stability (`@beta`, ...)

### DIFF
--- a/scripts/drupal/update-scaffold
+++ b/scripts/drupal/update-scaffold
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+CWD="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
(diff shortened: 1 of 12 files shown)


webflo:
:+1:

Issue #20, opened 2015-05-12
drush folder doesn't belong in "web" ?


yched:
... since it's not about web files ?

webflo:
Right, but i always thought needs the scope of a Drupal site. @weitzman created the PR for this feature.

@weitzman, @greg-1-anderson it there a easy solution to store the drush config outside of the webroot?

greg-1-anderson:
Really, the right solution is, as you said, to put vendor outside of the webroot.  Solve that problem, and Drush is outside of the Drupal root.

But I do agree with you.  My recommendation in my session put Drush inside the webroot as an unfortunate side-effect.  I probably should have addressed this explicitly in the session.

yched:
But vendor _is_ outside the webroot atm, right ? (the real one assembled by composer, not the one that is committed in core)

Also, just to be clear, this is only talking about the folder that holds "project-specific commandfiles, site aliases, and configuration", not the one with the drush/drush package and executables, which is already in the vendor outside the webroot :-)

greg-1-anderson:
Right.  I was thinking of a different project, for Drupal 7, that put vendor inside the docroot, and was thinking of vendor/drush/drush, not **WEBROOT**/drush, where the commandfiles go.

I opened https://github.com/drush-ops/drush/pull/1379 to consider **WEBROOT**/../drush as a default searchpath for Drush configuration and commandfiles.  If @weitzman agrees with this approach, this would allow us to put the drush folder at the project root, rather than in the web root.

webflo:
@greg-1-anderson Thanks for merging drush-ops/drush#1379 so quick. Just noticed that the `site-aliases` folder does not work either in /web/drush/sitea-aliases or /drush/site-aliases, Because [_drush_sitealias_find_alias_files](https://github.com/drush-ops/drush/blob/58320b79e0cbe2aa99e01ea2f03697bd42c44419/includes/sitealias.inc#L619) does not scan recursive.

greg-1-anderson:
Hm, maybe we should have merged that a little less quickly.  :)

Not to worry, we'll keep fixing it.  I think the best thing would be to make drush look for aliases in **CONF_DIR**/site-aliases in addition to **CONF_DIR**.

weitzman:
@greg-1-anderson - assuming you will take this on. i'm fine with whatever fix you prefer.

webflo:
[...]

PR #35, opened 2015-07-13
move drush commands folder outside webroot

yched:
PR for https://github.com/drupal-composer/drupal-project/issues/20

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,7 @@
             "web/modules/contrib/{$name}": ["type:drupal-module"],
             "web/profiles/contrib/{$name}": ["type:drupal-profile"],
             "web/themes/contrib/{$name}": ["type:drupal-theme"],
-            "drush/commands/{$name}": ["type:drupal-drush"]
+            "drush/contrib/{$name}": ["type:drupal-drush"]
         }
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,7 @@
             "web/modules/contrib/{$name}": ["type:drupal-module"],
             "web/profiles/contrib/{$name}": ["type:drupal-profile"],
             "web/themes/contrib/{$name}": ["type:drupal-theme"],
-            "web/drush/commands/{$name}": ["type:drupal-drush"]
+            "drush/commands/{$name}": ["type:drupal-drush"]
         }
     }
 }
--- a/drush/README.md
+++ b/drush/README.md
@@ -1,0 +1,1 @@
+The three subdirectories here hold project-specific commandfiles, site aliases, and configuration for Drush. See http://packages.drush.org/ for a directory of Drush commands installable via Composer.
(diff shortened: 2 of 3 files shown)


weitzman:
Actually we can keep the three subdirs and the README file now that Drush merged drush-ops/drush#1480

webflo:
Cool, https://github.com/drush-ops/drush/issues/1664 got it.

@yched Can you re-roll the patch and move drush to the repo root instead deleting it?

yched:
Like that ?

Also, should we recommend a separation of drush/commands/ into contrib/ & custom/, like we do for modules and themes ?
i.e. have `"drush/commands/contrib/{$name}": ["type:drupal-drush"]` in installer-paths ?

weitzman:
:+1: 

I think that separation is unnecessary. A project will have far fewer contrib commandfiles than contrib modules.

yched:
Sure, but the custom ones will be in a folder that is committed in the
project repo, while the contrib ones will be in a folder that is
composer-pulled & git-ignored ? So I'd think two folders still make sense
regardless of the # of projects in each ?

weitzman:
then the two folders should be 'local' and 'exernal'?

yched:
well, the "installer-path" already puts contrib modules & themes in
module/contrib & themes/contrib, not sure why we introduce a different
pattern here ?

weitzman:
I guess I dont have any real world experience here yet. I'll defer to you
guys.

webflo:
I think we should go with `"drush/contrib/{$name}": ["type:drupal-drush"]` for contrib and remove the three existing folders because most projects don´t need it? We could extend the README we some relevant links to documentation instead.

helderco:
I agree with @webflo on `"drush/contrib/{$name}"`.

yched:
I like having less folders in drush/ :-)

Where would a project put the things that were intended for drush/config and drush/site-aliases then ?

webflo:
@yched directly in `/drush`

yched:
Works for me. Will update the PR later today

yched:
done

webflo:
Thanks!: :+1:

yched:
Yay !

webflo:
Doh! The PR was incomplete. Fixed it with faac08e93b7bcdc74d660ec6b16a901f25755b4c

yched:
Er. Oops, sorry about that :-(

helderco:
Is this enough? drush isn't picking up my drush/drushrc.php.

```
$ composer create-project drupal-composer/drupal-project:8.x-dev some-dir --stability dev --no-interaction
$ cd some-dir
$ echo '<?php $options["shell-aliases"]["devel"] = "en -y devel";' > drush/drushrc.php
$ drush status
[no root, no config]
$ drush -r $PWD/web status
[with root, no config]
$ cd web
$ drush status
[with root, no config]
```

weitzman:
Put drushrc.php in the drush/ folder. It will get picked up anytime you specify --root ,or anytime your cwd is under web/

helderco:
@weitzman If you look at my commands above, I'm doing exactly that. Not working.

weitzman:
I can confirm that it isn't working. Maybe we added this for aliases only. In any case, this is a drush bug that I will fix this week.

helderco:
Yeah, I commented on drush-ops/drush#1379. Thanks.

weitzman:
Fixed by https://github.com/drush-ops/drush/commit/8a3620ee1e65bab7eb8069d57606db74148b8e64

PR #39, opened 2015-07-21
Ignore environment specific files

pfrenssen:
~~There are a bunch of defaults and example files in the `sites/` folder that should not be committed, such as `default.settings.php` etc. It would be good to add these to `.gitignore`.~~

I think we should also ignore `settings.php` and `services.yml`. These are not part of the standard code base but since they contain environment specific configuration and sensitive information (such as the database credentials) it would be good to  prevent this from being accidentally committed.


jonhattan:
At least `default.settings.php` and `default.services.yml` are required by Drupal installer.

See 
https://api.drupal.org/api/drupal/core!includes!install.core.inc/function/install_check_requirements/8

"The @drupal installer requires that the %default-file file not be modified in any way from the original download."

I'd prefer to preserve all default files. They are part of Drupal and handy to have them available when configuring environments, etc.

pfrenssen:
Won't these files be present after `composer install` is executed?

jonhattan:
`composer install` downloads `drupal/core` to `web/core`. See lines 20 and 33 of composer.json

webflo:
@pfrenssen `default.settings.php` and `default.services.yml` is not part of the `drupal/core` package.   The test passed because you did not remove them from the repo.

pfrenssen:
Ohh I was indeed not aware that these were not part of `drupal/core`, my apologies. Then let's only ignore the environment specific files.

webflo:
@pfrenssen Can you rebase your PR? I fixed the failing travis test in b8077a9.

greg-1-anderson:
Rather than ignoring settings.php, perhaps we should commit a placeholder settings.php here, and include + ignore a settings.local.php.  Something like the [Pantheon settings.php](https://github.com/pantheon-systems/drops-8/blob/master/sites/default/settings.php), minus the line that includes settings.pantheon.php.

This gives the user a place to commit common settings that should be shared across all environments. Environment-specific settings can go in settings.local.php.

The example .gitignore file shipped in Drupal 8 already ignores settings.*.php; this is a good standard to follow.

webflo:
I commit/track `settings.php` (without db credentials and salt) in all of my project. I am fine with this approach.

tstoeckler:
Yay, I also like to commit settings.php to the repo, but always thought there would be pushback against that approach. I have no problem with that, that would definitely allow solving #86 without hackery.

pfrenssen:
OK I agree, having a settings.local.php is a great solution.

pfrenssen:
Hmm `drush si` creates the settings file on installation and puts the salt and database credentials there. How are we going to deal with this?

greg-1-anderson:
What is the issue with settings.php?

When you run `composer create-project`, you inherit all of the committed files, e.g. settings.php, so there shouldn't be any issue with users committing modifications to settings.php.

If your issue is that you'd prefer these settings to go in settings.local.php, then the user can move them. Drupal writes this stuff, not Drush, so it would be inconvenient to make Drush auto-move them.

.gitignore
```diff
@@ -9,5 +9,9 @@ web/profiles/contrib
 # Ignore Drupal's file directory
 web/sites/default/files
 
+# Ignore environment specific settings
+web/sites/default/services.yml
```
deviantintegral, line 13: Do we really want to ignore services.yml? I've yet to see anything sensitive like API keys in it.

deviantintegral:
If we start doing much with committing / shipping settings.php, there's other potential pain points such as <a href="https://www.drupal.org/node/1232572">directory permissions</a>. If settings.php is shipped committed and modified, then `git commit -a` will include the database credentials, whereas if it's excluded you have to explicitly git add the settings file.

After reading this, I'm not sure if there's anything to do here? Seems like the two ignore lines aren't really needed.

.gitignore
pfrenssen, line 13: My reasoning was that it can be used to turn on things like Twig debugging which you don't want to have accidentally deployed to production.

pfrenssen:
Half a year on I don't care too much about this any more. We've adopted the settings.local.php practice. For me it's fine to close this. Maybe we can add a warning in the documentation somewhere, but I suppose people should just use common sense and not commit settings.php.

PR #45, opened 2015-08-08
Use bin/ for symlinked binaries

webflo:
I would like to add `bin` to the project template for binaries.

ping @pfrenssen, @yched, @greg-1-anderson


derhasi:
I would expect `bin` to only contain project specific commands, as therefore I personally would go with the default composer behaviour in `vendor/bin` for third party commands.

greg-1-anderson:
I usually set `"bin-dir": "bin"` for convenience, and put project-specific scripts in `scripts`.  For example, I do this with all of my projects based on [pantheon-systems/ci-scripts](https://github.com/pantheon-systems/ci-scripts), including [patheon-systems/example-drupal8-circle-composer](https://github.com/pantheon-systems/example-drupal8-circle-composer), which is based on this project, but adds behat testing on Circle CI and a Pantheon deploy step.

I kind of like it this way (with bin at the root), and I have also seen other projects do this.  However, I also add `./vendor/bin` to my $PATH by default, so that I can run third-party commands easily when my cwd is at a project root, so the default location isn't really less convenient--unless the project in question also relocates the `vendor` directory -- which I need to do in my example Drupal Composer projects order to deploy it to Pantheon.

For this project, I don't feel strongly about where bin should go.

bojanz:
I was just about to suggest the same thing, why was this issue closed?

greg-1-anderson:
This year I feel a little more strongly than last year that it is better to use the standard location, `vendor/bin`, unless the vendor directory is relocated.

One reason to keep the standard location is that php executables need to search for the vendor directory that they are associated with.  Most apps do tend to search for `../vendor`, but any that presumed they would always be in `vendor/bin` might not find their autoloader if bin is relocated.

I don't really know that this is a problem, and it is really a problem of composer, that allows for bin directory relocation without providing clear guidance on where it can go, and how applications should identify their autoloader location. All the same, I think it's best for this project to default to the standard location; folks can always relocate the bin directory themselves if that's what they want.

bojanz:
I've changed the drupalcommerce/project-base bin-dir to 'bin/', will see if it bites me.

greg-1-anderson:
@bojanz: I have done this lots, and never been bitten. The objection is mostly theoretical.

webflo:
Yeah both things work, but composer does not change the location of the binaries if you change it in the middle of the project. You have to reinstall all your packages.

webflo:
`composer exec` works good too.

PR #47, opened 2015-08-10
Use pwd instead of realpath

webflo:
`realpath` is not available OS X. I would like to have a update script that works on OS X.

The attached core works if we execute the script via `./scripts/drupal/update-scaffold`.

// @jonhattan What do you think?


jonhattan:
I can live with that :)

For the record, realpath is part of coreutils package in Debian jessie. In previous versions it was a package on its own. For OS X (of which I have no idea at all), google said https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man3/realpath.3.html

webflo:
Alright, Thanks for your feedback

PR #56, opened 2015-10-02
Update composer.json

webflo:


yched:
Why do we add things like mikey179/vfsStream, phpunit/phpunit ? aren't they require-dev provided by in drupal/core ?

jibran:
Yes all the required-dev can go away.

webflo:
`require-dev` is root-only config directive. https://getcomposer.org/doc/04-schema.md#require-dev

Its means we have to add these dependencies to your root composer.json

webflo:
Maybe we should provide them as a meta-package, but this fixes the tests until we have a better solution.

yched:
Aw. Thanks, didn't know that :-)

Issue #52, opened 2015-09-02
How to use drush?


michaellenahan:
When we are in the `web` directory, we can run `drush` like this:

```
cd web
../vendor/drush/drush/drush cr
```

Dumb question, but is there a better way to do it?

We can't currently add a symlink to the `web` directory, because the `drush` directory is already there.

```
cd web
ln -s ../vendor/drush/drush
ln: failed to create symbolic link ‘./drush’: File exists
```

rodrigoaguilera:
I personally use it on bash scripts
You can set a bash variable that holds the bin full path or instead of the variable alias drush and unalias it at the end.

derhasi:
Generally drush has a "redispatch" feature that can be enabled with setting `$options['drush-script']` to the path of your project specific drush. See https://github.com/drupal-composer/drupal-project/pull/16#issuecomment-93424497 for details. From what I understand you will need the latest drush version to be installed globally to make this work.
Maybe @greg-1-anderson can clarify it a little bit more.

For general use cases I created an experimental :boom: helper command to be installed globally: https://github.com/derhasi/buddy. With that small tool you could call drush via `b drush ...` from any place in your project, by simply defininy a `.buddy.yml`.

greg-1-anderson:
We are going to solve this with drush-ops/drush#1343; in particular, see my [comment on 18 May](https://github.com/drush-ops/drush/pull/1343#issuecomment-103192730) (I really need to get back to this issue....)

The general idea is that we will put this `drush finder` feature in Drush 7.x and Drush 8.x.  When you run `drush`, it will launch the `drush finder` in your $PATH (which is to say, it will run your global Drush).  This script in turn will notice that you have a site-local Drush associated with the site you have selected.  It will launch this Drush directly, rather than your global Drush.

webflo:
https://github.com/webflo/drushenv is one solution to the problem. Just run `echo "site" > .drush-version` in your project root and commit the .drush-version file. Drushenv will automatically dispatch all drush commands to the project specific drush installation.

greg-1-anderson:
drushenv is a good solution.  We are about to merge drush-ops/drush#1343 (later today, probably), so shortly you will be able to try that out.  Just install a version of Drush that contains the 'drush finder' script as your global Drush, and it will find and launch the correct site-local Drush for you.

So:

```
cd web
drush cr
```

[...]

PR #57, opened 2015-10-05
Use dev-master version of drush

bradjones1:
Now that https://github.com/drush-ops/drush/pull/1595 has landed, and per the comments in #52, the new "finder" stage for Drush is helpful/necessary to use the copy in the vendor directory. In my case, this works without any other changes or magic; I have a global /etc/drush/drushrc.php file that specifies the path to Drush, but this works inside the web/ directory, as well.

Also removed an unnecessary blank line.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,9 +18,8 @@
     "require": {
         "composer/installers": "^1.0.20",
         "drupal/core": "8.0.*",
-        "drush/drush": "8.*",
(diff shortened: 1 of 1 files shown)


webflo:
I tried the same in #56, but i have to look into the phpunit failures first.

webflo:
@bradjones1 Merged #56

PR #58, opened 2015-10-07
Update composer.json

janstoeckler:
This would be a welcome change if or once https://github.com/composer/installers/pull/264 is accepted, for using drupal-composer/drupal-project itself as a dependency.


webflo:
Afaik its not possible to use a composer project as template.

PR #59, opened 2015-10-07
Updates to update-scaffold script

bradjones1:
1. ...so you can choose a more arbitrary core version descriptor, e.g., for -dev. Includes README update.
2. Updates exclusion for sites, themes, modules so you don't end up clobbering data you may have there. Core doesn't really provide anything in those directories, anyway, so no need to expect upstream changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,10 +50,12 @@ Updating Drupal core is a two-step process.
 
 1. Update the version number of `drupal/core` in `composer.json`.
 1. Run `composer update drupal/core`.
-1. Run `./scripts/drupal/update-scaffold` to update files in the `web` directory.
-   This will update `web` with whatever the latest Drupal 8 release is. Review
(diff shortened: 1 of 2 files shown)


webflo:
Thanks for the PR. I think its useful to get updates for the default config files in sites (development.services.yml, example.settings.local.php etc) because its easier to discover new config options.

I wonder if we should remove the `--delete` flag or switch to explicit include options.

bradjones1:
It would be easy enough to carve in an exception for those, in the form of `default.*.*` ?

And this seems to work OK keeping the delete tag, so long as you keep the new exclusions... it is possible upstream Drupal does away with a scaffolding file (as is the case with `.csslintrc` in 8.1.x, for instance.)

webflo:
I merged it to prevent further data loss.

PR #69, opened 2015-11-15
Explicitly run post-install as a shell script

### LINKED ISSUE #65: post-install script fails to run on Windows
The composer post install command fails on Windows with the following error message:

```
'scripts' is not recognized as an internal or external command,
operable program or batch file.
Script scripts/composer/post-install.sh handling the post-install-cmd event returned with an error

  [RuntimeException]
  Error Output: 'scripts' is not recognized as an internal or external command,
  operable program or batch file.
```

I did a quick Github search: `filename:composer.json q:post-install-cmd*.sh` and noticed most projects run shell scripts with bash.

When I update the _scripts_ property in _composer.json_ the problem is resolved:

```
"scripts": {
  "post-install-cmd": "bash ./scripts/composer/post-install.sh"
}
```

bramcordie:
This fixes an issue that occurs when running the post-install script on Windows.
When you try to run the post-install script on Windows you receive this error:
'scripts' is not recognized as an internal or external command,
operable program or batch file.

Assuming you have the UNIX tools included with https://git-for-windows.github.io
installed, the post-install script will now run as expected.

Fixes #65

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
     "minimum-stability": "dev",
     "prefer-stable": true,
     "scripts": {
-        "post-install-cmd": "scripts/composer/post-install.sh"
+        "post-install-cmd": "sh ./scripts/composer/post-install.sh"
     },
     "extra": {
         "installer-paths": {


webflo:
Thanks!

PR #73, opened 2015-11-23
Add information about how to apply patches.

floretan:


webflo:
Thanks, but the URL is wrong. Can you fix it?

Issue #74, opened 2015-11-23
Update scaffold deletes profiles


jcnventura:
As with the modules and themes, the profiles page should be added to the exclude list of the rsync command.

webflo:
Fixed in #75

PR #75, opened 2015-11-23
Fix issue #74: exclude profiles from update-scaffold

jcnventura:


webflo:
Thanks!

Fixes #74

jcnventura:
That was fast :)

PR #77, opened 2015-11-24
Update scaffold files after 'composer update'.

greg-1-anderson:
This PR add pre-update and post-update scripts.  The pre-update script stashes the current Drupal version in a file; we put it in a dot file in the `vendor` directory, since that is already .gitignore'd.  The post-update script checks to see if the Drupal version changed after the update operation; if it has, it runs the update scaffold script.  This ensures that your composer-managed Drupal site will stay up-to-date, even if the scaffold files change.


jcnventura:
Neat! It would be nice to use the same logic for selecting the drush executable for the post-install command, as that one currently will depend on a local install or vendor/bin being in the path.

yched:
Shouldn't we delete the temp version file after post-update ? Feels weird to have file lying around that contains the current core version minus one ?

greg-1-anderson:
Good catch; I added a line to remove the stray version file after we recover its value.

I also updated the documentation in the README to match the new procedure.

greg-1-anderson:
The tests are passing now.

webflo:
Its a good idea in general, but i am not fully convinced on adding new shell script to the the repo. I experimented with robo (a php task runner) today and converted the existing update script to a task. (https://github.com/webflo/drupal-project/commit/e7b57dffa690c8d55cab2770fbcaf4df6d29a50b)

There is no need to use Drush to get the current version because everything shares the same autoloader. 

Some commands

<code>composer update-scaffold or composer robo update-scaffold</code>
<code>composer robo version:dump</code>

greg-1-anderson:
The RoboFile itself is a script, but since you can bundle multiple entrypoints into a single file, this could help reduce script file proliferation.  I integrated your suggestion with my PR; seems to work just fine, and there are definitely fewer script files now.

All of the tests are passing again; a little fiddling was required to get things to work in the CI environment.

derhasi:
We could also retrieve the the version information from the post-update event. And maybe provide that "Update Scaffold"-Feature as a separate Composer Plugin. See https://gist.github.com/derhasi/8afdd0ba314fa6953716 for a raw approach.

greg-1-anderson:
@derhasi: When using a Composer Plugin, do the pre-update events and post-update events execute as separate processes?  If they run in the same process, that would allow passing information from the pre-update event to the post-update event. However, if it is running in the same process, then the code probably won't see changes to drupal/core if the update operation swaps out all of the code.

I don't have time to test this right now. The version could be fetched via exec if it's a problem.

derhasi:
@greg-1-anderson, with the package update event, we get the package that is updated. We even get `initialPackage` and `targetPackage`. So we can only execute the code, when the package is `drupal/core`, otherwise no action would be performed. I guess that's all we need. Or am I missing something?

greg-1-anderson:
Oh, I understand now. We are getting the version number from the Composer metadata, rather than relying on the autoloader to access \Drupal::version.  That's much better; thanks.

I'll reroll using this technique as soon as I get a chance.

derhasi:
I'm currently trying to build a ComposerPlugin with that approach in https://github.com/derhasi/drupal-scaffold.

derhasi:
The plugin approach seems to work, see https://github.com/derhasi/drupal-project/commits/auto-scaffold. It still uses robo.

If you think, that approach is useful, I could move the plugin to `drupal-composer`. In addition we could make the plugin configurable with some "extra" options for overriding the excludes.

greg-1-anderson:
I think that this is a great idea.  I think that customization is critical; I would recommend an option to add additional excludes (the usual case), and another option to omit the default excludes (unusual).  With this addition, moving to drupal-composer would be a good thing to do.

Thanks for this. Happy to help out with config options later today if desired.

derhasi:
I moved it to https://github.com/drupal-composer/drupal-scaffold and it is now available via Packagist (https://packagist.org/packages/drupal-composer/drupal-scaffold)

greg-1-anderson:
drupal-composer/drupal-scaffold allows the scaffold update functionality to be added here in a very compact way.  Tests are all green; once again ready for review.

composer.json
```diff
@@ -28,7 +29,8 @@
         "jcalderonzumba/mink-phantomjs-driver": "dev-master#10d7c48c9a4129463052321b52450d98983c4332",
         "mikey179/vfsStream": "~1.2",
         "phpunit/phpunit": "~4.8",
-        "symfony/css-selector": "2.7.*"
+        "symfony/css-selector": "2.7.*",
+        "codegyre/robo": "^0.6.0"
```
yched: Not needed anymore as part of the root project ?

composer.json
```diff
@@ -19,7 +19,8 @@
         "composer/installers": "^1.0.20",
         "drupal/core": "8.0.*",
         "drush/drush": "dev-master",
-        "drupal/console": "dev-master"
+        "drupal/console": "dev-master",
+        "drupal-composer/drupal-scaffold": "dev-master"
```
yched: Nitpick, but since this is one of the base "drupal-composer" tools, could we move it up the list ?
The bottom of that list os probably going to be where each project adds its own contrib modules :-)

greg-1-anderson:
Cleaned up composer.json per suggestions from @yched.

webflo:
@greg-1-anderson The current code conflicts with our master branch. Can you resolve the conflict? Thanks!

greg-1-anderson:
Rebased and force-pushed.

greg-1-anderson:
I tried to make a separate PR based on this one to remove the scaffold files that we do not need any longer; however, I do not know how to submit it here. I could make a separate PR based from 8.x that included all of the changes here, but if I try to base it off of the branch in my fork, then the PR is submitted against my fork.

Perhaps we should just wait until this PR is merged; then I will submit another PR to remove the scaffold files.

weitzman:
Anything left to do on this one? @webflo?

webflo:
We have a few open PRs in the plugin itself (https://github.com/drupal-composer/drupal-scaffold/pulls). I think we should review/merge these first.

greg-1-anderson:
For what it is worth, I have a fork at [greg-1-anderson/drupal-project/pantheon-8.x](https://github.com/greg-1-anderson/drupal-project/tree/pantheon-8.x) that uses a branch of drupal-scaffold that is working fine.  It does not have any scaffold files checked in; it gets them all via drupal-scaffold, which is configured in composer.json to fetch from pantheon-systems/drops-8.

If you forked my fork, changed all occurrences of `code/` to `docroot/`, and deleted the drops-8 URL in the drupal-scaffold section of the `extra` section of composer.json, then this would likely work for you.

Hopefully we'll have all of the necessary changes committed to the official repositories shortly.

greg-1-anderson:
I added a couple of commits here to use the proposed strategy suggested in https://github.com/drupal-composer/drupal-scaffold/issues/9 and implemented in https://github.com/drupal-composer/drupal-scaffold/pull/20.  With this PR, drupal-scaffold only updates the scaffold files after `composer update`, or when explicitly requested. This project now calls `composer drupal-scaffold` in its post-install.sh script. It first checks to see if index.php already exists, and, if so, it presumes that all of the scaffold files have already been obtained. With this logic appearing directly in the post-install script, it is easy for derived projects to change the test to use whatever logic is best for their specific use-case, should the index.php test be insufficient.

Now, there are [no scaffold files checked in to this repository](https://github.com/drupal-composer/drupal-project/tree/334dba087636cd4d13686e858b082fc908f02f4b). The only references to the document root are in composer.json and scripts/composer/post-install.sh.

For almost all users, the default behavior of this PR should be exactly right. When the project is created the first time, there will be no index.php, so `composer drupal-scaffold` will run. If the user then commits all of the scaffold files (recommended), `composer install` will not run `drupal-scaffold` on the CI server. If the user does not check in composer.lock and the scaffold files (not recommended), then `composer install` will fetch the scaffold files every time.

scripts/composer/post-install.sh
```diff
@@ -1,21 +1,32 @@
 #!/bin/sh
 
+DOCUMENTROOT=web
+
+# Prepare the scaffold files if they are not already present
+if [ ! -f $DOCUMENTROOT/index.php ]
+  then
+    composer drupal-scaffold
+    mkdir -p $DOCUMENTROOT/modules
```
deviantintegral, line 9: Should these mkdir's be a part of the drupal-scaffold command? If I run `composer drupal-scaffold`, I think I'd expect the resulting web dir to be pretty much identical to what the core tarball ships.

scripts/composer/post-install.sh
```diff
@@ -1,21 +1,32 @@
 #!/bin/sh
 
+DOCUMENTROOT=web
```
deviantintegral, line 3: Do we want this to only be set if DOCUMENTROOT is empty, or instead be a command line option so it can be changed easily?

scripts/composer/post-install.sh
```diff
@@ -1,21 +1,32 @@
...
+fi
+
 # Prepare the settings file for installation
-if [ ! -f web/sites/default/settings.php ]
+if [ ! -f $DOCUMENTROOT/sites/default/settings.php ]
   then
-    cp web/sites/default/default.settings.php web/sites/default/settings.php
-    chmod 777 web/sites/default/settings.php
+    cp $DOCUMENTROOT/sites/default/default.settings.php $DOCUMENTROOT/sites/default/settings.php
+    chmod 777 $DOCUMENTROOT/sites/default/settings.php
```
deviantintegral, line 18: Add an echo here (or later below) alerting the user to the chmod? Otherwise, if someone commits without running an install they might accidentally deploy with insecure permissions.

scripts/composer/post-install.sh
greg-1-anderson, line 9: Sure; any part of this process that is completely invariant, such as creating the modules and themes directories & c., could become part of the drupal-scaffold command.

scripts/composer/post-install.sh
greg-1-anderson, line 3: I don't think that there is a good way to pass commandline variables to composer scripts (we expect this script to be run automatically via composer, not directly by the user).  We could change this to:

```
_DOCUMENTROOT=${DOCUMENTROOT-web}
```

Then, you could pass the document root value to the script via an environment variable:

```
DOCUMENTROOT=docroot composer update
```

However, I think this would be a bad idea, as it would cause havoc if you later ran composer update and did not specify the DOCUMENTROOT environment variable.

It would be better, I think, to ask Composer for the directory that `drupal/core` is installed to, and assume that the documet root is the parent directory of that.  Unfortunately, there is no composer command that will divulge that information.  We could potentially add a "tell me the installation location" command to drupal-scaffold, and use that, but it would require another entry in the `scripts` section to make it work.  This would, however, allow us to specify the document root only in composer.json, which would make it easier to customize. Customizing the document root in a committed file is the right way to do it, as you then are not at risk of messing up your update by running a command with the wrong parameters.

scripts/composer/post-install.sh
greg-1-anderson, line 18: Seems like a good idea to me.

greg-1-anderson:
@webflo @derhasi: Please take a look at the comments above, and weigh in.  It would be good if we could decide on our final strategy for [https://packagist.org/packages/drupal-composer/drupal-scaffold](https://packagist.org/packages/drupal-composer/drupal-scaffold), so that we can finish any remaining tasks and merge this PR.

bojanz:
Great work!

README.md
```diff
@@ -46,18 +48,15 @@ When installing the given `composer.json` some tasks are taken care of:
...
-   acceptable to Drush, e.g. `drupal-8.0.x` or `drupal-8.1.x`, corresponding to
-   the version you specified in `composer.json`. (Defaults to `drupal-8`, the
-   latest stable release.) Review the files for any changes and restore any
-   customizations to `.htaccess` or `robots.txt`.
+1. Run `git diff` to determine if any of the scaffolding files have changed. Review the files for any changes and restore any customizations to `.htaccess` or `robots.txt`.
 1. Commit everything all together in a single commit, so `web` will remain in
    sync with the `core` when checking out branches or running `git bisect`.
+1. In the event that there are non-trivial conflicts in step 2, you may wish to perform these steps on a branch, and use `git merge` to combine the updated core files with your customized files. This facilitates the use of a [three-way merge tool such as kdiff3](http://www.gitshah.com/2010/12/how-to-setup-kdiff-as-diff-tool-for-git.html). This setup is not necessary if your changes are simple; keeping all of your modifications at the beginning or end of the file is a good strategy to keep merges easy.
```
pfrenssen, line 59: Can we stick to 80 column width for the running text in the README? I'm asking this out of pure selfishness, then I can still read it in a half screen width terminal on my laptop ;)

scripts/composer/post-install.sh
pfrenssen, line 3: One way to do this is by allowing the user to save a settings file that contains the configuration and include this if it exists, something like this:

```
# Load config file if it exists.
if [ -f config.sh ]; then
  . config.sh
fi

if [ -z "$DOCUMENTROOT" ]; then
  DOCUMENTROOT=web
fi
```

But this might be too much, you can also simply provide a symlink if the the docroot needs to be somewhere else.

greg-1-anderson, line 3: The problem with supporting a config file on the script side is that there is no good way to get composer/installers to read the config value when setting the location of drupal/core in composer.json.

derhasi:
@greg-1-anderson; I merged your PR in drupal-scaffold, so you could change the branch.

For tasks regarding post-install.sh, before moving all functionality to drupal-scaffold (which could be an option), I think we could start with converting post-install.sh to a php script (e.g. `\DrupalComposer\DrupalProject::postInstall` or split it up).

This way we could retrieve the root (parent of drupal/core) via `$event->getComposer()`, similar to https://github.com/drupal-composer/drupal-scaffold/blob/master/src/Plugin.php#L76

In addition we could change the scripts section to use the `@drupal-scaffold` (see https://getcomposer.org/doc/articles/scripts.md#referencing-scripts) with the `post-create-project-cmd`-Event.

weitzman:
What are next steps here?

greg-1-anderson:
@webflo: Are you satisfied with the strategy and direction here? Would be good to get basic drupal-scaffold usage committed here.

webflo:
Yes, lets get it done. Can you do a rebase? Thanks!

greg-1-anderson:
Sure, I will rebase and do the light cleanup work today. I will put off moving functionality to drupal-scaffold for a later discussion, and focus on getting something we can commit here.

greg-1-anderson:
Please see #105 for the continuation.  I wanted to squash everything down to one clean commit based on the latest master, but did not want to --force push and erase the history of this branch, so I started a new PR.

PR #79, opened 2015-11-24
Use composer-patches

jcnventura:
Following the documentation work done in #73, I'm proposing to add cweagans/composer-patches to the required packages in composer.json, and to document how to use it in the README.md.

I'm pretty sure this package will be used by everyone as long as we have the current patch workflow in drupal.org.


greg-1-anderson:
Great idea: +1

webflo:
Yes it is a good idea. Can you squash both commits together? Thanks

jcnventura:
Squashed as requested.

webflo:
Thanks!

PR #82, opened 2015-11-28
settings.php and services.yml need not be executable.

5sw:
Giving everything the 777 permissions works, because it basically disables all permission checks. That doesn't make it right though.

Changed the script to not add the executable bit to settings.php and services.yml.

Those permissions still are insecure though. For development on a local machine this is probably ok, but for deployment to a server more permissions than 440 or maybe 640 (if they are not owned by the user the web server runs as) are not acceptable.

### DIFF
--- a/scripts/composer/post-install.sh
+++ b/scripts/composer/post-install.sh
@@ -4,14 +4,14 @@
 if [ ! -f web/sites/default/settings.php ]
   then
     cp web/sites/default/default.settings.php web/sites/default/settings.php
-    chmod 777 web/sites/default/settings.php
+    chmod 666 web/sites/default/settings.php
 fi
 
 # Prepare the services file for installation
 if [ ! -f web/sites/default/services.yml ]
   then
     cp web/sites/default/default.services.yml web/sites/default/services.yml
-    chmod 777 web/sites/default/services.yml
+    chmod 666 web/sites/default/services.yml
 fi
 
 # Prepare the files directory for installation


jcnventura:
I'd suggest 640, as the file has just been copied from the default, and for sure, at least the settings.php will need to be edited.

webflo:
I think the permissions for settings.php are not relevant because Drupal makes settings.php readonly anyway. (`system_requirements()`around line 370)

bcremer:
Why not respect the users umask when setting the permission? The umask is also repected when new files are created.

```
chmod $(printf '%o' $((0666 & ~$(umask))))  web/sites/default/settings.php
```

This will set the permissions depending on the users umask setting.

```
$ umask 0 && chmod $(printf '%o' $((0666 & ~$(umask)))) web/sites/default/settings.php && ls -l web/sites/default/settings.php 
-rw-rw-rw- web/sites/default/settings.php

$ umask 077 && chmod $(printf '%o' $((0666 & ~$(umask)))) web/sites/default/settings.php && ls -l web/sites/default/settings.php 
-rw------- web/sites/default/settings.php

$ umask 022 && chmod $(printf '%o' $((0666 & ~$(umask)))) web/sites/default/settings.php && ls -l web/sites/default/settings.php 
-rw-r--r-- web/sites/default/settings.php

$ umask 027 && chmod $(printf '%o' $((0666 & ~$(umask)))) web/sites/default/settings.php && ls -l web/sites/default/settings.php 
-rw-r----- web/sites/default/settings.php
```

This script depends on `#!/bin/bash`. Not sure if that will also work with `#!/bin/sh`.

See: http://unix.stackexchange.com/a/249856/537

pfrenssen:
Even the files folder is given world write access. This is worse than the settings.php file, since the permissions of settings.php are checked by Drupal and if these are writable it is reported to the admins.

webflo:
Done in #132

pfrenssen:
This is not fixed, even in #132 the access to the settings files and files folder are still to permissive.

PR #84, opened 2015-12-05
Use stable versions of Drush and Drupal Console.

greg-1-anderson:
Both Drush and Drupal Composer now have stable releases on their 8.x branches.  We should prefer these over the dev releases.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,8 +19,8 @@
         "composer/installers": "^1.0.20",
         "cweagans/composer-patches": "~1.0",
         "drupal/core": "8.0.*",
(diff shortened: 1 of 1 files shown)


derhasi:
``` json
{
  "drush/drush": "^8.0.0",
  "drupal/console": "^0.9.9"
}
```

should work.

Edit: drupal/console follows a different versioniong: https://github.com/hechoendrupal/DrupalConsole

webflo:
Merged in #87

webflo:
Thanks!

PR #85, opened 2015-12-05
Add drupal-security-advisories and roave/security-advisories.

greg-1-anderson:
webflo/drupal-security-advisories and roave/security-advisories are super cool, and very helpful should folks get into the habit of pinning their requirements to very specific versions.  Seems like it would be a good idea to encourage folks to use them by default.

Maybe we should move webflo/drupal-security-advisories  to drupal-composer/drupal-security-advisories first?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
     "require": {
         "composer/installers": "^1.0.20",
         "cweagans/composer-patches": "~1.0",
-        "webflo/drupal-security-advisories": "8.0.x-dev",
+        "drupal-composer/drupal-security-advisories": "8.0.x-dev",
         "roave/security-advisories": "dev-master",
         "drupal/core": "8.0.*",
         "drush/drush": "~8.0",
--- a/composer.json
+++ b/composer.json
@@ -18,6 +18,8 @@
     "require": {
         "composer/installers": "^1.0.20",
         "cweagans/composer-patches": "~1.0",
+        "webflo/drupal-security-advisories": "8.0.x-dev",
+        "roave/security-advisories": "dev-master",
         "drupal/core": "8.0.*",
         "drush/drush": "~8.0",
         "drupal/console": "~0.9"


derhasi:
I like this one.

@webflo do you want to move it to drupal-composer or does that conflict with Circle CI?

webflo:
I should work under a different namespace as well. I move the package tonight.

webflo:
Done, triggered a Circle CI build to update the package names.

greg-1-anderson:
Tests are failing with:

```
Exception: You must provide a SIMPLETEST_BASE_URL environment variable to run some PHPUnit based functional tests.
```

Perhaps there is another test group we need to exclude; however, I do not get this failure when I run the tests locally.

greg-1-anderson:
Excluding simpletest avoided the failure. I don't know if you want to fix it up so that the simpletests can run; I don't like fiddling with simpletest, really.

bojanz:
I guess we just need a rebase now?

greg-1-anderson:
Rebased the PR; tests now pass again w/out removing simpletest. Thanks.

bojanz:
Woohoo! 

ship it

weitzman:
:+1:

webflo:
Do you actually use it in development? I had the impression it causes a significant slow down during the install/update process. We should look at the D7 version to get some real life data, because the conflict file for D8 is rather small in comparison to D7.

weitzman:
When D8 file gets big, we can provide an option to bypass it? Seems reasonable to commit this now and deal with slowness in a year or two.

greg-1-anderson:
```
if ($slow && ($securityNotImportant || $trustAllMyDependenciesToNeverPinVersions)) {
   removeSecurityAdvisories("$myproject/composer.json");
}
```

Or something like that. We could include this in the template, and adjust the README to advise folks to remove it if they don't want it.

I think this should also go into the 7.x branch.  Been meaning to time how long it takes there.

weitzman:
Needs reroll. But first, do folks still want to see this happen here? Makes sense to me.

webflo:
I think its a good idea in general, but it slows down the install and update process. People are already ranting about the performance of composer. I don't what to make it worse by adding it by default. Especially because this project is a starting point for many people and they may get a false impression and don't know that is making it so slow.

greg-1-anderson:
I wonder if there would be some easy way to do this via a composer script; something like a `composer require` command that does not actually update the compser.json file.

Maybe a short script that:
- copied the composer.json file to a temporary location
- ran `composer require --working-dir` with the security advisory packages
- report yes or no & remove temporary location

A lot of the benefit would be lost by making this a manual rather than automatic step, though.  I wonder if this might be a good feature request to submit to VersionEye -- have a config file that could do extra steps e.g. `composer require` prior to analysis.  They already claim to do security checks for php.  It's unclear to me where they draw their data from, but their service might be a good alternative to a composer-based solution.

webflo:
They use the https://security.sensiolabs.org/database according to the blog post https://blog.versioneye.com/2015/05/14/php-security-vulnerabilities/. We would need a client side implementation of SensioLabs security checker ...

webflo:
I won't add it as a default. Sorry. PRs for docs welcome.

PR #91, opened 2015-12-21
Drupal in Docker on OSX

andyworsley:
As per README.md update.

Comments welcome.


derhasi:
I think this is out of scope for this project. Providing some best practices for Docker should be handled in a spearate project similar to [DrupalVM](http://www.drupalvm.com/). There are already some around on github.

PR #98, opened 2016-01-09
Add web/drush.wrapper scaffold file for restricting commands/aliases/config to local repo.

weitzman:
Drush has support for a drush.wrapper in Drupal root which customizes how Drush gets called. The wrapper I propose here adds --local to all Drush requests. This tells Drush not to scan for config, site aliases, and commandfiles in global locations (e.g. ~/.drush, /etc/drush, ...). This environment isolation is very helpful - all Drush requests run the same no matter who's system you are on. Furthermore, it emphasizes that Drush files go in this repo's /drush dir, not in legacy places like ~/.drush. In the rare cases when globally installed commands are needed and can't be copied to our /drush for some reason, user can add --no-local to her request (or in the extreme permanently remove --local from this file).

In order to test this PR, make sure you have some commandfiles, config, or aliases installed to ~/.drush. Then run a `drush sa --root=/path/to/web`. Notice how your global commands/aliases/config are no longer in effect with this PR.

### DIFF
--- a/web/drush.wrapper
+++ b/web/drush.wrapper
@@ -1,0 +1,32 @@
+#!/usr/bin/env sh
+#
+# DRUSH WRAPPER
+#
+# A wrapper script which launches the Drush that is in your project's /vendor
+# directory.  Copy it to the root of your project and edit as desired.
+# You may rename this script to 'drush', if doing so does not cause a conflict
+# (e.g. with a folder __ROOT__/drush).
+#
+# Below are options which you might want to add. More info at
+# `drush topic core-global-options`:
+#
+# --local       Only discover commandfiles/site aliases/config that are
+#               inside your project dir.
+# --alias-path  A list of directories where Drush will search for site
+#               alias files.
+# --config      A list of paths to config files
+# --include     A list of directories to search for commandfiles.
+#
+# Note that it is recommended to use --local when using a drush
+# wrapper script.
+#
+# See the 'drush' script in the Drush installation root (../drush) for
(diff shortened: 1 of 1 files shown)


weitzman:
ping @greg-1-anderson for review.

greg-1-anderson:
I've started to use this feature, and think it still needs a little smoothing out, per https://github.com/drush-ops/drush/issues/2023. Some easy option to add on the commandline to negate the use of the drush wrapper (opposite effect as --local) would also be good.

That said, there will be a drush.wrapper equivalent to this one in the Pantheon version of this project.

webflo:
https://github.com/drush-ops/drush/issues/2023 is still open. I think we should adopt the script without the `--local` flag. This solves the main use case and we don't have to cd back and forth between project root and web.

weitzman:
I'm fine with committing this without --local. I wrote above about the value of this option. I will say that clients who really want to bring in drush commands/config from outside the project can still use drush's --config and --alias-path.

weitzman:
I removed --local. I think this is ready. webflo proposed something similar in #202 (either script would be fine with me).

weitzman:
Actually, this should go into project root not /web. Lets close this in favor of #202 (or ask people to use drupal-shim project).

yched:
@weitzman Just curious : what is the "drupal-shim project" ?

weitzman:
Sorry, I mistyped. Here it is: [drush-shim](https://github.com/webflo/drush-shim)

yched:
Ah, right :-)

PR #101, opened 2016-01-21
Ignore drupal/drupal and all Drupal 6 and 7 Modules

webflo:
Similar to #71, but we exclude D6 and D7 modules this time.


webflo:
Committed in 10d1c80.

PR #104, opened 2016-01-26
Fixes #102: include settings.local.php from default.settings.php.

### LINKED ISSUE #102: Uncomment default.settings.php local.settings.php area
default.settings.php comes with the following lines commented out:

```
# if (file_exists(__DIR__ . '/settings.local.php')) {
#   include __DIR__ . '/settings.local.php';
# }
```

These should be uncommented by default going forward. Acquia hosting does this by default and now more and more drupal projects would like to use settings.php as a configuration piece for non-sensitive information while db connection info is kept in local.settings.php

This also may require edits to .gitignore. I haven't provided a pr for this but am more than willing once we have a discussion/ it's decided it's a good direction for the project.

allgood2386:
I did a quick search through the project, I don't believe this requires an update to .gitignore, this doesn't appear to be contained in drupal core anymore.

### DIFF
--- a/web/sites/default/default.settings.php
+++ b/web/sites/default/default.settings.php
@@ -709,6 +709,6 @@ $settings['container_yamls'][] = __DIR__ . '/services.yml';
  *
  * Keep this code block at the end of this file to take full effect.
  */
-# if (file_exists(__DIR__ . '/settings.local.php')) {
-#   include __DIR__ . '/settings.local.php';
-# }
+ if (file_exists(__DIR__ . '/settings.local.php')) {
+   include __DIR__ . '/settings.local.php';
+ }


allgood2386:
Strange, I'm not sure why this failed only php56 dev. I've turned on travis on my fork and will look into this.

allgood2386:
So I took a look through this and nothing I did that I can think of should have failed this on composer install like the single test did. It passed on all tests on my personal fork which has the same travis file and all other builds in this repo. 

Seems to me this was a composer issue at the time of install but I'm super new to travis so I could be incorrect here.

greg-1-anderson:
https://github.com/drupal-composer/drupal-project/pull/77 deletes this file.

greg-1-anderson:
Drupal 8.1.x-dev still has default.settings.php as near as I can tell.  See: http://cgit.drupalcode.org/drupal/tree/sites/default

Is there an uncommitted issue that proposes to remove it?

allgood2386:
So the way the project is set up, or I thought, per discussions on #drupal in irc w/ @webflo , is that the project contains it's own default.settings.php which right now is a duplicate of core. I propose that the project modify the lines for local settings, and uncomment it by default, so that settings.php is treated as public configuration and local settings would contain secret that you don't want to commit. If I'm not incorrect acquia already does this on d7 and I assume d8, and I think most developers using this project (again from conversation in irc, so this might require more discussion) would like to be using the local settings in this way.

allgood2386:
Also for what it's worth, this does not affect ../vendor/bin/drupal site:install when the local file is present. I can update the documentation, seems like it's more confusing than I initially thought?

webflo:
I would like to have this feature, but we have to revaluate how this works after #105 went in.

allgood2386:
I've never tried drupal-scaffold so I'll have to catch up but from a quick read, couldn't we just use that, and define our own source? I'm referring to this section about the extras area. https://github.com/drupal-composer/drupal-scaffold#configuration

greg-1-anderson:
So, default.setting.php is now pulled from the upstream Drupal. We could customize it in the post-drupal-scaffold-cmd script after https://github.com/drupal-composer/drupal-scaffold/pull/24 is merged.

It is not necessary to customize the default options for drupal-scaffold, as it already copies default.settings.php by default.

derhasi:
I do not think the "default.settings.php" should be altered, as this is the default core file. I'd rather put such functionality in post-install.sh, where the settings.php is initially copied.

greg-1-anderson:
Seems reasonable. The danger in not altering default.settings.php, though, is that users who follow the Drupal 8 reinstall instructions will copy default.settings.php over their settings.php file as recommended. They will probably remember to put back any customization that they put in themselves, but are less likely to catch things done automatically for them.

A better solution would be to fix the Drupal 8 reinstall instructions, but I don't know that we have a good avenue forward there.

allgood2386:
Does anyone think this should involve a drupal.org issue? What would the arguments against having the altered default.settings.php in core? Have core include the if( settings.local.php) {} lines by default?

I guess you could say if someone got write access to sites/default it makes it a bit easier to inject some php, but that's all I can really think of at this moment.

If we think it's a good idea, I'd say we put that issue in drupal.org issue queue and for now go with one of the proposed solutions, knowing they are temporary?

greg-1-anderson:
Including a file that does not exist is innocuous; I am not sure why these lines are not included by default in core. Perhaps someone is concerned about the performance impact of a single file_exists? (shrug)

allgood2386:
I created an issue on d.org if anyone cares to argue for or against the proposed change. I do think this belongs on their end. Drupal site install should do this by default imo. https://www.drupal.org/node/2676538

I suppose this pr should be closed as it's already out of date.

PR #105, opened 2016-01-29
Use drupal-composer/drupal-scaffold

greg-1-anderson:
Remove the 'web' directory, and fetch it using drupal-composer/drupal-scaffold.  Drupal-scaffold also handles scaffold updates.

Continuation of #77.


webflo:
Thanks for the hard work! This is a awesome feature. Rebased it after #106 went in.

PR #106, opened 2016-02-05
Workaround for permission denied when executed from Git Bash

fmeirelesfon:
When executed from a Git Bash environment the command `mkdir -m777` fails with a `Permission denied` error. Executing the mkdir followed by chmod solves the problem.

permission denied


webflo:
Thanks for the fix. I am glad it works on windows, i had some serious doubts :)

PR #113, opened 2016-02-20
install destination for third-party libraries

iainhouston:
Third-party libraries (e.g. juicebox.js which drupal/juicebox needs) currently, and somewhat to my surprise in Drupal 8, have to be installed in `web/sites/all/libraries`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -47,6 +47,7 @@
             "web/modules/contrib/{$name}": ["type:drupal-module"],
             "web/profiles/contrib/{$name}": ["type:drupal-profile"],
             "web/themes/contrib/{$name}": ["type:drupal-theme"],
+            "web/sites/all/libraries/{$name}": ["type:drupal-library"],
             "drush/contrib/{$name}": ["type:drupal-drush"]
         }
     }


weitzman:
I think best practice will be to add bower/grunt/gulp step to your build for js and css stuff. The line here is pretty harmless, but I dont expect it to become best practice. I could be wrong though.

rodrigoaguilera:
I added that line to my composer.json because it eases the migration from drush make . Is also recommended here https://www.drupal.org/node/2471553

Is nice that for example the colorbox module can declare a dependency for the js library it gets downloaded where the colorbox module is expecting it.

kasperg:
Take https://www.drupal.org/node/2471553 with a grain of salt.

In my opinion the big difference here is that the other types are defined by the project on http://packagist.drupal-composer.org. The name/type `drupal-library` is entirely up to the individual project as it have to define such packages itself. Libraries will not get the type automatically.

If we choose to standardize on `drupal-library` when I think we should document the purpose of it within this project.

I do not think we can rely 100% on other package managers like bower or npm to solve this. In this case I could not find the Juicebox library using either system. Also I agree that having the library end up in the location whether the module expects it is a good thing.

webflo:
https://github.com/oomphinc/composer-installers-extender helps with that. It provides and installer for type:library packages. I think the default path for type:drupal-library should web/libraries

Nebel54:
@iainhouston The sites/all/ folder structure is not recommended anymore, as https://api.drupal.org/api/drupal/sites%21README.txt/8.2.x states:

> It is now recommended to place your custom and downloaded extensions in the
> /modules, /themes, and /profiles directories located in the Drupal root. The
> sites/all/ subdirectory structure, which was recommended in previous versions
> of Drupal, is still supported.

 So i would guess we should put custom libraries directly into web/libraries instead.

webflo:
I went ahead with `web/libraries``

PR #114, opened 2016-02-21
Create a hash_salt value for settings.php in post-install.sh

greg-1-anderson:
In case user wishes to copy an existing database rather than install a new site.

I put this in with `sed` so that the hash salt value would be placed in settings.php in the same place it would if the Drupal installer were run.  Perhaps the extra complexity is not worth the benefit, though; not sure how I feel about this.


rodrigoaguilera:
I think this is interesting for when there is no installation happening

weitzman:
IMO this is too much magic. Chances are that other settings.php edits are needed anyway.

AlexSkrypnyk:

PR #121, opened 2016-03-09
Improve directory scaffolding; add .gitkeep files

bradjones1:
Use a foreach loop, and also add a .gitkeep file to the directories.

**Motivation** - I tried to run tests using `run-tests.sh` and got some vague failures since Drupal's test script isn't particularly verbose. Tracked this down to `core/tests/bootstrap.php` which was trying to recursively search the `profiles` directory, which did not exist. That's because I had originally created this project on another machine, committed it to git, and then fired it back up on my current node. However, the empty directories created in `post-install.sh` did not get committed, per git's usual behavior.

The directories aren't re-created on `composer install` because `web/autoload.php` already existed. Empty directory, can't scan, makes PHPUnit unhappy.

### DIFF
--- a/scripts/composer/post-install.sh
+++ b/scripts/composer/post-install.sh
@@ -6,9 +6,11 @@ DOCUMENTROOT=web
 if [ ! -f $DOCUMENTROOT/autoload.php ]
   then
     composer drupal-scaffold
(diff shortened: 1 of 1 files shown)


webflo:
Thanks, rebased the change in a0fd501.

PR #122, opened 2016-03-09
Use drupal-drupal instead of drupal-core.

jibran:
After this change `drupal-scaffold` is borken but this will allow us to not copy the core from drupal.org again after composer install.


webflo:
I appreciate the idea, but there are few issues.
- drupal/drupal is not updated after every change, packagist pulls new information every few days
- drupal/drupal does not expose its composer dependencies anymore
- New .gitignore is complex, it should include .htaccess and a few others

jibran:
Thank you for appreciating the idea.

> drupal/drupal is not updated after every change, packagist pulls new information every few days

For stable project this is fine but for core developer yes this is an issue.

> drupal/drupal does not expose its composer dependencies anymore

Does this mean that when we'll remove the vendor directory form drupal core. It will not work properly? 

> New .gitignore is complex, it should include .htaccess and a few others 

For all my stable projects .gitignore is always complex. I added a rebots.txt example. I can add .htaccess as well but last two commits can go to follow up PR. If we agree to work on first two commits.

Anyway this is just POC on my behalf if you don't agree feel free to disregard it. :)

derhasi:
>  but this will allow us to not copy the core from drupal.org again after composer install.

I do not see what problem using "drupal/drupal" solves. Can you explain that a little bit further?

In addition, with using drupal/drupal you will have a lot more problems. For example every time drupal/drupal will be updated it will erase your whole modules and themes folders and any changes you made in the autoload.php, robots.txt, and so on.

jibran:
> I do not see what problem using "drupal/drupal" solves. 

This is an official drupal composer package. Not some alternate package https://github.com/drupal-composer/drupal-core. AFAIK https://github.com/drupal-composer/drupal-core is created because there was no composer file in to drupal root directory and now we have a composer.json in root directory courtesy `wikimedia/composer-merge-plugin` then why not leverage that instead. Once the vendor directory is removed form the core we'll have only one vendor directory per drupal project.

> For example every time drupal/drupal will be updated it will erase your whole modules and themes folders and any changes you made in the autoload.php, robots.txt, and so on.

Yes I'm aware of that and that's what I meant when I said `drupal-scaffold` is broken. We have to change `drupal-scaffold` according to above requirements which would be to take backup of custom and default directories before install/update of `drupal/drupal` and restore them after the install/update and hack autoload.php for core. And we assume that in an ideal workflow these files will all be tracked in version control so you can always get them back.

derhasi:
> This is an official drupal composer package. Not some alternate package https://github.com/drupal-composer/drupal-core.

That's not true. `drupal/core` is authorized by Dries: see https://packagist.org/packages/drupal/core. It was created to support the new layout in Drupal 8 to separate core functionality from project specific code, hence the `core` directory in the first place.

`drupal-scaffold` is created to support a workflow that still relies on some of the original package files. But in an evolving project you will get rid of most of them anyway.

btw: if you want to go with drupal/drupal in your project using https://github.com/derhasi/composer-preserve-paths could be helpful.

jibran:
> That's not true. drupal/core is authorized by Dries: see https://packagist.org/packages/drupal/core. It was created to support the new layout in Drupal 8 to separate core functionality from project specific code, hence the core directory in the first place.

I definitely think we should reconsider this now because we have `composer.json` in the root directory and removing the vendor from root is a [go](https://www.drupal.org/node/1475510#comment-10958035).

Thank you for https://github.com/derhasi/composer-preserve-paths reference I'll definitely have a look at it.

jibran:
Anyway, I had a really good discussion thank you @webflo and @derhasi. A lot of the things are clearer to me now.

PR #127, opened 2016-03-17
Fixing line endings in post-install script.

dsdobrzynski:
Discovered in issue #124

PR #128, opened 2016-03-18
Issue #127: Fix post-install script

webflo:
This should fix the failing create-project command. Composer changed the behaviour in https://github.com/composer/composer/commit/55b0ed8c8be74af348d9fec6702da3efd50feb46


greg-1-anderson:
I'm a little bit confused about the reference to the composer commit above. I just ran `composer create-project`, and it still works as well as it always did. Did Composer behavior really change at all here?

In any event, Composer changing behavior or not changing behavior doesn't really matter--I am just curious for my own edification. As far as this PR is concerned, switching the post-install script from a bash script to equivalent PHP code means that we are no longer dependent on `composer` being available in the user's $PATH at post-install time, so this seems like a good enhancement to me.

+1.

webflo:
@greg-1-anderson You can spot the change in [create-project-command.test](https://github.com/composer/composer/commit/55b0ed8c8be74af348d9fec6702da3efd50feb46#diff-c1506ae77a3a87c8ff8479560f5181d8L10) easily.

scripts/composer/ScriptHandler.php
```diff
@@ -0,0 +1,65 @@
...
+
+namespace DrupalProject\composer;
+
+use Composer\Script\Event;
+use Symfony\Component\Filesystem\Filesystem;
+
+class ScriptHandler {
+
+  protected static function getDrupalRoot($project_root) {
+    return $project_root .  '/web';
```
jibran, line 16: I think we should get this directory(`/web`) form installer paths in composer.json.

scripts/composer/ScriptHandler.php
```diff
@@ -0,0 +1,65 @@
...
+
+    // Required for unit testing
+    foreach ($dirs as $dir) {
+      if (!$fs->exists($root . '/'. $dir)) {
+        $fs->mkdir($root . '/'. $dir);
+        $fs->touch($root . '/'. $dir . '/.gitkeep');
+      }
+    }
+
+    // Prepare the settings file for installation
```
jibran, line 44: I think we should allow user to turn copying file on or off using some variable in `composer.json`. If you like the idea we can create a new issue for this.

scripts/composer/ScriptHandler.php
webflo, line 16: Yeah drupal-scaffold has this feature already. I think we could detect it via composer installers path or add something link "drupal-root-dir" to the composer.json. A own subclass of ScriptHandler should work as well ...

PR #132, opened 2016-03-23
Fixed settings.php, services.yml and files directory creation process

ModulesUnraveled:
The settings.php and services.yml files were getting created empty with the permissions "--w--wx-wT". They need to use the octal "0666" to work correctly.

The files directory wasn't being created at all (a copy and paste error was trying to create another services.yml file.) Also, the directory was being created with incorrect permissions due to umask. I've added two lines to temporarily set umask to "0" and then restore the original value after the files directory is created.

### DIFF
--- a/scripts/composer/ScriptHandler.php
+++ b/scripts/composer/ScriptHandler.php
@@ -44,21 +44,23 @@ class ScriptHandler {
     // Prepare the settings file for installation
     if (!$fs->exists($root . '/sites/default/settings.php')) {
       $fs->copy($root . '/sites/default/default.settings.php', $root . '/sites/default/settings.php');
-      $fs->chmod($root . '/sites/default/settings.php', 666);
+      $fs->chmod($root . '/sites/default/settings.php', 0666);
       $event->getIO()->write("Create a sites/default/settings.php file with chmod 666");
     }
 
     // Prepare the services file for installation
     if (!$fs->exists($root . '/sites/default/services.yml')) {
       $fs->copy($root . '/sites/default/default.services.yml', $root . '/sites/default/services.yml');
-      $fs->chmod($root . '/sites/default/services.yml', 666);
+      $fs->chmod($root . '/sites/default/services.yml', 0666);
       $event->getIO()->write("Create a sites/default/services.yml file with chmod 666");
     }
 
     // Prepare the files directory for installation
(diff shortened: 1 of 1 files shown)


scripts/composer/ScriptHandler.php
```diff
@@ -44,21 +44,23 @@ public static function createRequiredFiles(Event $event) {
     // Prepare the settings file for installation
     if (!$fs->exists($root . '/sites/default/settings.php')) {
       $fs->copy($root . '/sites/default/default.settings.php', $root . '/sites/default/settings.php');
-      $fs->chmod($root . '/sites/default/settings.php', 666);
+      $fs->chmod($root . '/sites/default/settings.php', 0666);
       $event->getIO()->write("Create a sites/default/settings.php file with chmod 666");
```
webflo, line 48: Maybe we should update the comments as well?

scripts/composer/ScriptHandler.php
```diff
@@ -44,21 +44,23 @@ public static function createRequiredFiles(Event $event) {
...
+      $fs->chmod($root . '/sites/default/services.yml', 0666);
       $event->getIO()->write("Create a sites/default/services.yml file with chmod 666");
     }
 
     // Prepare the files directory for installation
     if (!$fs->exists($root . '/sites/default/files')) {
-      $fs->chmod($root . '/sites/default/services.yml', 777);
-      $event->getIO()->write("Create a sites/default/files directory with chmod 777");
+      $oldmask = umask(0);
+      $fs->mkdir($root . '/sites/default/files', 0777);
```
webflo, line 61: `mkdir` takes chmod as second argument. https://github.com/symfony/filesystem/blob/d46a5ec18b78bdd0ab04798d6710ca91c3588837/Filesystem.php#L88

webflo:
Thanks for catching this :)

ModulesUnraveled:
No problem. I'm about to give a talk in New Orleans about site-building in D8, so I'm going through to make sure everything works as expected. :)

scripts/composer/ScriptHandler.php
ModulesUnraveled, line 48: I updated the comment to say "// Create the files directory with chmod 0777". Does that work?

webflo:
Pushed to 8.x.Thanks!

PR #138, opened 2016-04-14
Disable 3 Drush pm-* commands as they conflict with Composer maintenance.

weitzman:


greg-1-anderson:
This is a great idea. In addition to giving drupal-project folks a heads-up if they try to use a pm-\* command when they should be using `composer update` et. al., it also provides a discoverable policy file that can be customized for their own purposes.

Should the message say what file the message is coming from, for the interested?  Or maybe this should be a Drush feature -- if a validate hook fails, also include a message showing the user the commandfile that caused it.

derhasi:
Great feature.

Is it possible to grab the demanded projects, so we can print some additional message like:

>  Use 'composer require "drupal/pathauto:8._" "drupal/devel:8._"' instead.

greg-1-anderson:
The recommendation is to use `composer update` instead of `drush pm-update` or `drush pm-updatecode`, and `composer require` instead of `drush pm-download`. There is no need for any contrib modules.

Perhaps the message could be clarified in this regard.

weitzman:
If someone has better text for the error message, feel free to post it here. I think this can go in as is.

greg-1-anderson:
Perhaps instead of "This codebase is managed via composer commands, not drush pm commands", use:

> Please use `composer update` and `composer install` in preference to `drush pm-update` and `drush pm-download`.

I agree that this could just go in as it is, though.

weitzman:
Updated error message as suggested. This is ready, IMO.

webflo:
The error message needs work.

PR #139, opened 2016-04-19
Support Drupal 8.1 and newer.

bartfeenstra:
This allows sites to install any compatible Drupal 8 release instead of just 8.0.x.


webflo:
Thanks!

PR #159, opened 2016-06-06
Use packages.drupal.org repository

oxyc:
Or is it still bleeding edge?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
     "repositories": [
         {
             "type": "composer",
-            "url": "https://packagist.drupal-composer.org"
+            "url": "https://packages.drupal.org/8"
         }
     ],
     "require": {
@@ -32,7 +32,7 @@
         "phpunit/phpunit": "~4.8",
         "symfony/css-selector": "~2.8"
     },
-    "conflict": {
+    "replace": {
         "drupal/drupal": "*"
     },
     "minimum-stability": "dev",


ryanaslett:
its not _totally_ bleeding edge, but theres a couple of rough patches here n there.

geerlingguy:
Platform.sh [just announced](https://platform.sh/2016/05/drupal-8-starter-kit-now-with-php-7-and-composer/) they're defaulting to the new repository, so it seems like it may be stable enough for people to start using it in the wild (and helping smooth the rough patches!).

ryanaslett:
Im probably being overly critical of the _theoretical_ rough edges.   The main outstanding issue is this: https://www.drupal.org/node/2744539, predicated by this: https://www.drupal.org/node/2744455.

So even though its only one module, there's about 17 other modules with a reference to media_entity as a dependency, if we extract that out further to see how many modules depend on media_entity through some strange dependency chain, it leaves just enough room for me to call it a 'rough edge'

shorock:
Working well for me on D8, and noticeably faster.

Thinking D7, the packages.drupal.org/7 endpoint doesn't provide the synthesized (for lack of a better term) drupal/drupal:7.\* jsons that packagist.drupal-composer.org does.  Seems to provide just about everything else for 7. I haven't been following the discussion enough to know if that's in the works for the drupal.org endpoints.

For interim testing in anticipation of a better answer, in my composer-built D7 projects, I've been doing...

```
  "repositories": [{
    "type": "composer",
    "url": "https://packages.drupal.org/7"
  },
  {
    "type": "composer",
    "url": "https://packagist.drupal-composer.org"
  }],
```

to prefer the drupal.org but fall back.

ryanaslett:
@shorock , could you elaborate on the d7 issue? According to the logfiles, I havent seen anybody even _use_ the d7 composer endpoint yet, so Im curious if something is just broken or what.

drupal/drupal should be found at packagist.org, so as to prevent chicken/egg scenarios where `composer create-project drupal/drupal` doesnt know to go to the packages.drupal.org endpoint.

fafnirical:
One breaking change in switching to the Drupal.org Composer Service is that distributions and install profiles aren't yet supported: https://www.drupal.org/node/2715505

shorock:
@ryanaslett - I was guessing that I was pretty early on it, but I've been building things in the @islandora world (includes a whole darned lot of github submodules, very nice to use composer) using the p.d.o/7 endpoint for a couple weeks now.

The packagist.org/packages/drupal/drupal doesn't have 7.\* builds.  The deprecating drupal-composer packagist does.

Anyway, just mentioning the 7.x branch in passing. Don't let it block the 8.x PR, though both the noted concerns above are worth considering.

webflo:
Related issue https://github.com/drupal-composer/drupal-packagist/issues/61

shorock:
Just an update for those who don't do D7 Composer that quite a bit has changed (positively) in a fortnight.  The packages.drupal.org/7 endpoint now returns drupal/drupal:7.\* packages (although, interestingly, only to 7.43, not 7.44... will check/report there).  Packagist.org/packages/drupal/drupal also now has the 7.\* packages that it didn't have two weeks back.

ryanaslett:
Not sure why 7.44 wasn't there, the timestamp of the file coincided with the release. I've re-ran the metadata builder and 7.44 is there now. I'll open an issue to remind myself to investigate if there is some sort of race condition happening on new releases.

ryanaslett:
Also, I dont think that d7 drupal/drupal exists on packagist: https://packagist.org/p/drupal/drupal%2466ae3a5c3958ab51c9a19bdc5a36b45f55d1ae2a0e82857e79ad8855a5f3822b.json  is the current provider file, and a `composer info drupal/drupal --all` without the d.o. repo only shows the 8.x versions.

webflo:
@oxyc I think we could keep the `drupal/drupal` conflict. Because d.o switched to `drupal/core` as base package. Or we remove it completely?

webflo:
Thanks! Committed to 8.x in https://github.com/drupal-composer/drupal-project/commit/49184d678ff0de7f0cda0b597a7b170c6e5261e6

PR #161, opened 2016-06-08
Adds a name key to repositories

ryanaslett:
Having a key name in the repositories will make it marginally easier to change the repo from the command line:

`composer config repositories.drupal composer https://packages.drupal.org/8`

is easier than telling folks to edit json.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,12 +9,12 @@
             "role": ""
         }
     ],
-    "repositories": [
+    "repositories": {
         "drupal": {
             "type": "composer",
-            "url": "https://packagist.drupal-composer.org"
+            "url": "https://packages.drupal.org/8"
         }
-    ],
+    },
     "require": {
         "composer/installers": "^1.0.20",
         "drupal-composer/drupal-scaffold": "^2.0.1",
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
         }
     ],
     "repositories": [
-        {
+        "drupal": {
             "type": "composer",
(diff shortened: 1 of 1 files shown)


fafnirical:
That's not valid JSON.

From the examples on the [composer.json schema ](https://getcomposer.org/doc/04-schema.md#repositories) docs, it looks like you can't name repositories, just packages.

ryanaslett:
Whoops, you're correct about the json being invalid. 

But as far as the schema goes, you can name repositories, thats exactly what happens when you execute the composer config command above. And the composer-schema spec says that a repository can be either an object or an array: https://github.com/composer/composer/blob/master/res/composer-schema.json#L333

fafnirical:
@ryanaslett Oh, huh. The docs do say:

> **Note:** Order is significant here. When looking for a package, Composer will look from the first to the last repository, and pick the first match. By default Packagist is added last which means that custom repositories can override packages from it.

However, composer/composer#2802 suggests that it's not documented because it'd be confusing (which it is, since JS objects don't have a sense of order).

ryanaslett:
ah. good find. Honestly, I'd rather see https://github.com/drupal-composer/drupal-project/pull/159 get in sooner but I think there's still a couple of outstanding things that prevent a wholesale switchover, so this will make it easier to provide transisiton instructions.  The bigger issue for me is the 108 forks of this project that all point at drupal-packagist.

webflo:
I will go with #159 instead of this one.

PR #168, opened 2016-06-21
[7.x] Better ScriptHandler.php, migrate from drustack/drustack-standard

hswong3i:
Migrate from https://github.com/drustack/drustack-standard/blob/master/src/Composer/ScriptHandler.php, include:
- Original script handlers, clone from Drupal 8.x Drupal\Core\Composer\Composer
- Generate info meta data into .info, logic clone from drush 8.x

### DIFF
--- a/scripts/composer/ScriptHandler.php
+++ b/scripts/composer/ScriptHandler.php
@@ -23,11 +23,55 @@ use Symfony\Component\Process\Process;
 class ScriptHandler
 {
     protected static $packageToCleanup = [
-        'behat/mink' => '/(driver-testsuite|tests)$/',
+        'behat/mink' => '/(tests|driver-testsuite)$/',
+        'behat/mink-browserkit-driver' => '/(tests)$/',
+        'behat/mink-goutte-driver' => '/(tests)$/',
+        'doctrine/cache' => '/(tests)$/',
+        'doctrine/collections' => '/(tests)$/',
+        'doctrine/common' => '/(tests)$/',
+        'doctrine/inflector' => '/(tests)$/',
+        'doctrine/instantiator' => '/(tests)$/',
         'egulias/email-validator' => '/(documentation|tests)$/',
+        'fabpot/goutte' => '/(Goutte\/Tests)$/',
+        'guzzlehttp/promises' => '/(tests)$/',
+        'guzzlehttp/psr7' => '/(tests)$/',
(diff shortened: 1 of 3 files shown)


hswong3i:
Some update due to https://www.drupal.org/node/2760467 changes.

scripts/composer/ScriptHandler.php
```diff
@@ -0,0 +1,445 @@
...
+        $root = getcwd().'/web';
+
+        // Prepare the settings file for installation.
+        if ($fs->exists($root.'/sites/default/default.settings.php')
+            && !$fs->exists($root.'/sites/default/settings.php')) {
+            $fs->copy(
+                $root.'/sites/default/default.settings.php',
+                $root.'/sites/default/settings.php'
+            );
+            $fs->chmod($root.'/sites/default/settings.php', 0666);
```
madalinignisca, line 242: Why?

scripts/composer/ScriptHandler.php
```diff
@@ -0,0 +1,445 @@
...
+        }
+
+        // Prepare the services file for installation.
+        if ($fs->exists($root.'/sites/default/default.services.yml')
+            && !$fs->exists($root.'/sites/default/services.yml')) {
+            $fs->copy(
+                $root.'/sites/default/default.services.yml',
+                $root.'/sites/default/services.yml'
+            );
+            $fs->chmod($root.'/sites/default/services.yml', 0666);
```
madalinignisca, line 253: Why?

scripts/composer/ScriptHandler.php
```diff
@@ -0,0 +1,445 @@
...
+            $event->getIO()->write('Create a sites/default/services.yml file with chmod 0666');
+        }
+
+        // Create the files directory with chmod 0777.
+        if (!$fs->exists($root.'/sites/default/files')) {
+            $oldmask = umask(0);
+            $fs->mkdir($root.'/sites/default/files', 0777);
```
madalinignisca, line 260: Really why???

madalinignisca (commented):
Why should permissions be set so open? Permissions responsibility should be an important part before deploying live and not automated like this.

jcnventura:
I've been syncing the 7.x branch with 8.x. Other than the vendor cleanup stuff, I'm not sure why all the code that is in drustack/drustack-standard but not in drupal-composer/drupal-project is needed. Then again, I don't see the reason to even cleanup anything.

PR #170, opened 2016-06-25
Check the Composer version before downloading the dependencies

pfrenssen:
I experienced a problem yesterday during a workshop where people weren't able to scaffold Drupal correctly. This turned out to be caused by doing an initial installation using an outdated version of composer.

We do not have the `composer.lock` file committed (to avoid merge conflicts downstream) and an initial installation without a lock file is now considered to be an update command. This change has been recently introduced in Composer (in https://github.com/composer/composer/pull/5035). If someone is using an older version this means the Drupal scaffolding is not performed.

Since this change has been introduced shortly before Composer 1.0.0 I would propose to check for this version.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,6 +44,12 @@
     },
     "scripts": {
         "drupal-scaffold": "DrupalComposer\\DrupalScaffold\\Plugin::scaffold",
+        "pre-install-cmd": [
+            "DrupalProject\\composer\\ScriptHandler::checkComposerVersion"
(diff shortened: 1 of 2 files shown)


webflo:
Can we add a doc block with the link to the GitHub issue? Thanks!

pfrenssen:
Sure! I've added a docblock.

webflo:
Merged, thanks!

PR #177, opened 2016-07-15
Fix whitespace in indents and other places

hussainweb:
I know this is a really small change but my eye kept hurting with the wrong indentation in composer.json. I fixed that and couple of other whitespace issues I found.

It's a quick fix. :)


webflo:
Thanks!

Issue #176, opened 2016-07-14
Add option to move "core" directory above of web root.


stevector:
Today's Drupal contrib security fixes reminded me of the inherent danger in keeping PHP files within the docroot. https://twitter.com/drupalsecurity/status/753263548458004480 This project already makes a big improvement by moving the vendor directory above the docroot. The modules, themes, profiles and even core could eventually be moved as well.

There is an old issue on Drupal.org on the topic: https://www.drupal.org/node/1672986 Without movement in that issue, not much can be done here but I at least wanted to open this issue here to gauge interest.

fago:
Klausi documented a way of doing that, see https://www.drupal.org/node/2767907 and https://twitter.com/_klausi_/status/754740752295616513. Would be great if the composer setup could handle this by default?

webflo:
👍 I think we can do it as a composer script.

webflo:
Please review https://github.com/drupal-composer/drupal-project/pull/180.

michaellenahan:
I want to review #180, but I would like to check I'm doing it the right way.

```
git clone git@github.com:webflo/drupal-project.git my-project
cd my-project
git checkout paranoia-mode
composer install
```
This seems to work - I get an ```app``` directory and the ```web``` directory has no php files - but I would like to check if that's the recommended way to test this PR. Thanks!

webflo:
Yeah this is correct. I am currently rewriting it as a composer plugin to provide its as additional command.

webflo:
Actually i am not sure if we should write it a full-blown composer plugin because, we have drush or console could act as task runner since they require a local installation anyway.

michaellenahan:
(irrelevant question deleted)

jkribeiro:
Please, see https://github.com/drupal-composer/drupal-project/pull/180#issuecomment-347977101

jkribeiro:
I've created an experimental Composer plugin based on the PR https://github.com/drupal-composer/drupal-project/pull/335

https://github.com/jkribeiro/drupal-composer-paranoia

I think withe a Composer Plugin we can decouple the functionality from `drupal-project` turning the "paranoia mode" optional.
 
We can move the Plugin to `drupal-composer` org if necessary.

normanlolx:
Seems https://github.com/drupal-composer/drupal-paranoia fixed it.

PR #180, opened 2016-07-18
Issue #176: Implement paranoia mode

webflo:


rainer-friederich:
What about fonts from custom themes? Perhaps we should make the file types that should get copied configurable.

webflo:
Yes, thats the idea. We can add it to the composer.json. I think we should have the ability to link whole folders as well. E.g. Symlink the css and js folder of a certain theme instead of individual files.

composer.json
```diff
@@ -39,31 +39,37 @@
...
+            "DrupalProject\\composer\\AssetInstaller::install"
         ]
     },
     "extra": {
         "installer-paths": {
-            "web/core": ["type:drupal-core"],
-            "web/modules/contrib/{$name}": ["type:drupal-module"],
-            "web/profiles/contrib/{$name}": ["type:drupal-profile"],
-            "web/themes/contrib/{$name}": ["type:drupal-theme"],
+            "app/core": ["type:drupal-core"],
```
klausi: why do we need the "app" directory? Shouldn't core be at the root? This just one annoying directory level deeper with no benefit.

webflo: Not necessarily, but this project had a folder which contained modules, profiles, theme and some dotfiles before. And it was not on the repo root before.

heddn: I appreciate breaking out all the application stuff into a separate folder. I can then put things like patternlab directly into the app folder and sync over what I need.

.gitignore
```diff
@@ -1,13 +1,14 @@
 # Ignore directories generated by Composer
 drush/contrib
 vendor
-web/core
-web/modules/contrib
-web/themes/contrib
-web/profiles/contrib
+app/core
```
fago: bikeshed: app is very generic, why not just use "drupal" ? In case you have some other stuff like the mentioned pattern-lab in the repo also, that makes things clear.

composer.json
fago: I agree that "app" does not make so much sense, as "vendor" would have to be part of the app also?

scripts/composer/AssetInstaller.php
```diff
@@ -0,0 +1,86 @@
...
+    $fs = new Filesystem();
+    $fs->symlink(realpath($extra['drupal-app-dir']) . '/sites/default/files', $extra['drupal-web-dir'] . '/sites/default/files');
+  }
+
+  public static function createSymlinks($appDir, $webDir) {
+    $finder = new Finder();
+    $cfs = new \Composer\Util\Filesystem();
+
+    $names = [
+      '.htaccess',
```
fago: hm, I guess some scaffold files like the webserver config should be just _moved_ to the web dir. There is no point in having them in the app-dir.

scripts/composer/AssetInstaller.php
```diff
@@ -0,0 +1,86 @@
...
+      '*.js',
+      '*.png',
+      '*.svg'
+    ];
+
+    $finder = $finder->ignoreDotFiles(FALSE)->in($appDir);
+    foreach ($names as $name) {
+      $finder->name($name);
+    }
+    $finder->exclude('sites/default/files');
```
fago: We should make those configurable also.

fago: I've been wondering whether it makes sense to add robo for that kind of things?

composer.json
ostark: "/app" is the common folder in symfony and laravel for code you own. While "/vendor" is the default place for 3rd party (composer) dependencies.  This structure is common pattern these days:

/app
/vendor
/var (or /storage)
/web (or /public)

scripts/composer/AssetInstaller.php
```diff
@@ -0,0 +1,86 @@
...
+    static::createStubPhpFile($extra['drupal-app-dir'], $extra['drupal-web-dir'], 'index.php');
+    static::createStubPhpFile($extra['drupal-app-dir'], $extra['drupal-web-dir'], 'core/install.php');
+    static::createStubPhpFile($extra['drupal-app-dir'], $extra['drupal-web-dir'], 'core/rebuild.php');
+
+    // Optional?
+    static::createStubPhpFile($extra['drupal-app-dir'], $extra['drupal-web-dir'], 'core/modules/statistics/statistics.php');
+
+    // Symlink public files
+    $fs = new Filesystem();
+    $fs->symlink(realpath($extra['drupal-app-dir']) . '/sites/default/files', $extra['drupal-web-dir'] . '/sites/default/files');
```
jkribeiro: IMO is a good thing create the files symlink using relative paths, absolute symbolic links don't work when the filesystem is mounted elsewhere.
`$fs->symlink('../../../' . $extra['drupal-web-dir'] . '/sites/default/files', $extra['drupal-app-dir'] . '/sites/default/files');`

jkribeiro:
@webflo could you please update this PR to solve those files conflicts

webflo:
@jkribeiro Did a rebase against 8.x

medienverbinder:
Whenever possible I use the approach from [drupal.org](https://www.drupal.org/node/2767907) for new installations and I think a working "paranoia mode" is very useful. What is going to happen next with this issue? Can I do something to move forward?

scripts/composer/AssetInstaller.php
```diff
@@ -0,0 +1,87 @@
...
+  public static function install(Event $event) {
+    $extra = $event->getComposer()->getPackage()->getExtra();
+    if (!isset($extra['drupal-app-dir'])) {
+      throw new \RuntimeException('Please configure drupal-app-dir in your composer.json');
+    }
+    if (!isset($extra['drupal-web-dir'])) {
+      throw new \RuntimeException('Please configure drupal-web-dir in your composer.json');
+    }
+    foreach (static::$frontControllers as $fileName) {
+      static::createStubPhpFile($extra['drupal-app-dir'], $extra['drupal-web-dir'], $fileName);
```
jkribeiro, line 43: It seems that `makePathRelative()` expects absolute paths to do the comparison. In this case it should be:
`static::createStubPhpFile(realpath($extra['drupal-app-dir']), realpath($extra['drupal-web-dir']), $fileName);`

jkribeiro, line 43: Before call `createStubPhpFile()`, the web and app folders must exist, otherwise the process will return an error
```
[Symfony\Component\Filesystem\Exception\IOException]  
  Unable to write to the "/" directory.
```

This was not happening before because `createSymlinks()` was being called before `createStubPhpFile()` and creating the folders.
https://github.com/drupal-composer/drupal-project/pull/180/commits/b111239718037aad21693492b035f9a3ee3758cf#diff-a97cd408fd45330f7510ba6d71ab4907R25

Suggestion:
```
// Ensure that the app and web directories exist.
$cfs = new \Composer\Util\Filesystem();
$cfs->ensureDirectoryExists($web_dir);
$cfs->ensureDirectoryExists($app_dir);

// Call createStubPhpFile().

//  Call createSymlinks().
```

jkribeiro:
From: https://github.com/drupal-composer/drupal-project/pull/335

Initially, I know that is not a good thing to have parallel PR for the same change, I'm gonna explain:
I've been running the paranoia mode on some of my projects for a while and since https://github.com/drupal-composer/drupal-project/pull/180 has been without activity for some time, I decided to fork it and create this https://github.com/drupal-composer/drupal-project/pull/335 with the changes that I've made. I tried to create a PR from https://github.com/drupal-composer/drupal-project/pull/180 but it is conflicting, so, here is a fresh one from 8.x.

__Main changes made from the original PR__
- __Recreate the web dir on every assets installer process__: The web directory is disposable, I think its needs to be deleted and recreated from scratch on every installation to remove the old assets and install the new ones.
- __Added new config for Public files folder location__: A new *optional* config was added to be able to set a custom symlink target for the the public files at the web dir. This could help with custom locations on different servers. The new optional config is `drupal-web-dir-public-files`.

I'm running the paranoia mode on Acquia Cloud servers, to be able to customize this for Acquia structure, I've created https://github.com/jkribeiro/drupal-project-acquia-paranoia-mode for additional installation/instruction.

Issue #192, opened 2016-09-14
How to use contributed Drush commands


ModulesUnraveled:
I was talking with @greg-1-anderson the other day and asked how I could use a set of drush commands across projects, and allow others to use them too. He said I should file an issue here, so that's what I'm doing. :)
## Scenario:

I have a set of [useful Drush shell-aliases](https://github.com/ModulesUnraveled/Drush-Shell-Aliases) that make Drupal 8 Development easier. For example: `drush @mysite.loc confex` will
1. Disable a list of specified development modules
2. Then run config-export
3. Then re-enable those same development modules

This is so that their configuration is never stored through Configuration Management as it would cause issues on live environments etc. (ie: You don't have Devel on live, but the core.extensions file says it should be enabled, so a config-import will fail.)

I'd like to be able to `composer-require` those aliases across my projects, as well as any others that people come up with.

https://github.com/drupal-composer/drupal-project/tree/8.x/drush says

> See http://packages.drush.org/ for a directory of Drush commands installable via Composer.

But that page doesn't exist, so I'm not sure how to create my own, or where to get existing ones.
## Current solution

I've added a drushrc.php in the drush directory with the following content:

```
<?php

 /**
 * Load additional drushrc.php files
 */
if (file_exists(__DIR__ . '/contrib/drush-shell-aliases/drushrc.php')) {
  include __DIR__ . '/contrib/drush-shell-aliases/drushrc.php';
}
```

Then, I can run `composer require "modules-unraveled/drush-shell-aliases:^1.0"`

and since the composer.json file has:

```
    "extra": {
        "installer-paths": {
            "drush/contrib/{$name}": ["type:drupal-drush"]
        }
    }
```

I just set my set of shell-aliases to be type "drupal-drush" and they get installed into the location specified in the drushrc.php file above ^.
## The Problem

This works pretty well, but there are a number of fragile assumptions being made.
1. I have to create the initial drushrc.php file that includes the other one
2. I'm only including one specific contrib set, so if I find another (or create additional unrelated ones) they won't get included unless I also add another require to the top-level drushrc.php
3. Probably other stuff I haven't thought about yet
## The Question _(finally)_

Would it be possible to include a default drushrc.php that checks for other drushrc.php files in "drush/contrib/*" directories (recursively?) so that they are all included automatically as soon as they are required through composer?

I think that would be super cool!

Thanks!

greg-1-anderson:
You can search for Drush extensions on Packagist via:

[...]

PR #195, opened 2016-09-23
#192: Update Readme with a new link

zaporylie:
Link to Drush commands directory is invalid. Should be https://packagist.org/search/?type=drupal-drush


webflo:
Thanks!

PR #194, opened 2016-09-20
using drupal.org official repositories

madalinignisca:
I've altered the skeleton to use the drupal.org repository. For me it worked well.

PR #198, opened 2016-09-25
Update composer.json to use drupal.org packagist.

janstoeckler:
We might want to start using https://packages.drupal.org/8.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
     "repositories": [
         {
             "type": "composer",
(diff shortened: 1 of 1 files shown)


webflo:
We have two open PRs. I am going to merge it shortly after DrupalCon.

D8: https://github.com/drupal-composer/drupal-project/pull/159
D7: https://github.com/drupal-composer/drupal-project/pull/194

PR #206, opened 2016-10-11
Update README.md composer require command

keopx:
With the lastest version change composer require versions system.


webflo:
Thanks!

PR #207, opened 2016-10-12
Ignore files generated by DrupalScaffold.

ojacquet:
It is useless to commit the files generated by DrupalScaffold since they are overwritten after a composer install. However, there should be a way to apply patches to them so that the build process can apply changes to those files after a build.


ojacquet:
composer install doesn't overwrite those files. Only composer update and then you can see the changes and commit anyway. So, not sure if this makes sense.

ojacquet:
Related to:

https://github.com/drupal-composer/drupal-scaffold/issues/9
https://github.com/drupal-composer/drupal-scaffold/pull/20

PR #209, opened 2016-10-14
WIP feat(docroot): Use a dedicated docroot with no core files in it

klausi:
Trying to use this project with an empty docroot for security reasons, see https://www.drupal.org/node/2767907

This is not complete yet. 2 problems:
1) we need additional web server config directives to map static assets (js, css, images). We want to avoid the rsync hack or copying around files, a web server is perfectly capable of mapping the files for us.
2) The generated settings.php file needs `$settings['file_public_base_url']` which contains a domain that we just set to localhost for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,16 +39,36 @@ all files not excluded by the .gitignore file.
 
 When installing the given `composer.json` some tasks are taken care of:
 
-* Drupal will be installed in the `web`-directory.
-* Autoloader is implemented to use the generated composer autoloader in `vendor/autoload.php`,
-  instead of the one provided by Drupal (`web/vendor/autoload.php`).
-* Modules (packages of type `drupal-module`) will be placed in `web/modules/contrib/`
-* Theme (packages of type `drupal-theme`) will be placed in `web/themes/contrib/`
-* Profiles (packages of type `drupal-profile`) will be placed in `web/profiles/contrib/`
+* Drupal will be installed in the project root directory.
+* A dedicated document root "web" directory will be created. This is where your
+* web server configuration should point to. See the web server section below.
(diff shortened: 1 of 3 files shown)


webflo:
How compares this approach with #180?

klausi:
In contrast to #180 we avoid using one more artificial "app" directory and bring drupal directly to the project root. Which is perfectly fine in Drupal 8, you directly have your modules and themes folder there ==> nice.

fago:
hm, I'm not sure this is much better than the old approach with improved webserver config? I mean, this is just going from a webserver config blacklist to a whitelist. This is better, but still requires a proper webserver config to work correctly. But compared to the separate docroot, it's not so obvious what gets served and what not, - what means it can still happen that one accidentally serves unwanted files like html,css and JS assets from test directories?

klausi:
*.html is not in the whitelist, so that will never be served. Yes, you might still be serving some unwanted JS/CSS or images, but a lot less of course than having all the Drupal source code files in the docroot.

If you want to stop serving stuff from tests directories then you can improve the regex in the web server config. We can totally do that! (BTW: I have not figured out the nginx config yet, because we need to switch the docroot).

This points yet again to the problem in Drupal that modules and themes need to mark their assets that should be served. We don't have a mechanism for that yet. We could just have a simple convention that each module has an "assets" or "web" folder. That will not happen in the Drupal 8 release cycle, so we need to live with this hack.

IMO the main advantage of this is that we avoid yet another "app" subdirectory people have to navigate.

stephenpurkiss:
Just wondered if there was anything that could be done to move this forward, sounds like a Good Thing!

webflo:
We have a plugin for this. https://github.com/drupal-composer/drupal-paranoia

PR #210, opened 2016-10-15
Use common standards

Saphyel:
Because all the ignore paths are folders, is a good practice to add at the end `/` so IDE/s etc.. are aware that is a folder rather than a file.


smalot:
Paths shouldn't start by a "/" too ?

webflo:
Thanks!

webflo:
Added the / at the beginning.

PR #211, opened 2016-10-26
Change drupal-scaffold command name

gergelypap:
New command name could be `drupal:scaffold`, as Composer separates these commands in different sections. Think of it as a command namespace.

Example:
Imgur

Of course these would comply to [drupal-scaffold](https://github.com/drupal-composer/drupal-scaffold) too.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,7 @@
         ]
     },
     "scripts": {
-        "drupal-scaffold": "DrupalComposer\\DrupalScaffold\\Plugin::scaffold",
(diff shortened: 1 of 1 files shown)


webflo:
I like the idea, not sure if i am going to merge it because of BC reasons.

derhasi:
I don't think BC is a real reason here, as drupal-project is a one-off starting point and not something you constantly would merge back. In the latter case you would see the merge conflict anyway.

jcnventura:
I think this is obsoleted by #275.

PR #215, opened 2016-10-31
Add documentation on how to install libraries using composer.json

hussainweb:
The composer.json in the project allows placing components marked as drupal-library in the
correct directory, but most libraries are not available to composer by default. We should
add documentation on how to add a repository with an example to help get users started.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,3 +105,40 @@ section of composer.json:
     }
 }
 ```
+
+### How can I add libraries using composer.json?
+
+It is possible to use libraries with composer with an extra repository for each
+library you want to use. For example, to use colorbox:
+```json
+"repositories": [
+    ...,
+    {
+        "type": "package",
+        "package": {
+            "name": "jackmoore/colorbox",
+            "type": "drupal-library",
+            "version": "1.6.4",
+            "dist": {
+                "url": "https://github.com/jackmoore/colorbox/archive/1.6.4.zip",
+                "type": "zip"
+            },
+            "source": {
+                "url": "https://github.com/jackmoore/colorbox.git",
+                "type": "git",
+                "reference": "tags/1.6.4"
+            }
+        }
+    }
+],
+```
+
+The version has to be manually updated with a new release as composer can't
+detect a new version from the above package.
+
+When managing libraries with composer this way, you may not want to add it to
+version control. In that case, add specific directories to the .gitignore file.
+```
+# Specific libraries (which we manage with composer)
+docroot/libraries/colorbox
+```
--- a/README.md
+++ b/README.md
@@ -142,3 +142,12 @@ version control. In that case, add specific directories to the .gitignore file.
 # Specific libraries (which we manage with composer)
 docroot/libraries/colorbox
 ```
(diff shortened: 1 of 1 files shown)


webflo:
I think inline package definitions are not a best practice. We should at least mention the limitations. See the blue box on https://getcomposer.org/doc/05-repositories.md#package-2

hussainweb:
It's a fair point. Even though there are limitations, it is still essential in many cases. You may use bower or npm for your theme but that approach doesn't really work well here where this needs to be present site-wide in libraries directory. I have added the warning with a link back to getcomposer.org.

weitzman:
I ma finding more and more js libraries available via Composer. Maybe we should mention those as well (use installer-paths to place them in web/libraries)

kasperg:
Another options is [Asset Packagist](https://asset-packagist.org/) which makes Bower and NPM available through Composer.

I like the idea there are currently some caveats regarding placement that I am not too keen on - hiqdev/asset-packagist#20. Perhaps they can be addressed.

joelpittet:
Trying to figure out how to deal with deleting `test` folders that come with packages https://www.drupal.org/node/1189632
Indicated on this project:
https://www.drupal.org/project/dropzonejs

webflo:
@joelpittet You could try to creata a subclass <code>\Drupal\Core\Composer\Composer</code> and override <code>\Drupal\Core\Composer\Composer::$packageToCleanup</code>. Add add the subclass to your composer.json as additional script.

joelpittet:
Toying with who will get the responsibility, possibly `npm/yarn`, but maybe composer, thanks for the suggestion @webflo

rodrigoaguilera:
I found a way that can make the installation of external libraries from npmjs.org or bower into the folder /web/libraries/ easier

Recently this composer repo was created
https://asset-packagist.org/
but there still the problem that this libraries do not set a "type" for "composer/installers" to use so is difficult to set a path like /web/libraries/$name

There is this plugin that add a default type to all packages
https://github.com/oomphinc/composer-installers-extender

So we modify this project to add the assets repo, the plugin and write something like the following in the composer.json
```json
"installer-types": ["library"],
"installer-paths": {
  "web/libraries/{$name}": [
    "type:drupal-library",
    "vendor:npm-asset",
    "vendor:bower-asset"	
  ],
}
```
So we can do things like
composer require "npm-asset/jquery.easing":"~1.0"

and it will place the library into /web/libraries/jquery.easing

There is another explanation here
https://asset-packagist.org/site/about

No need to duplicate js libraries in packagist for a drupal "version".
In projects with a lot of external libraries the repositories section can get quite messy and https://github.com/fxpio/composer-asset-plugin is slow and complicated.
Also overcomes the limitations referenced in this PR about changing the library version.


Is it interesting to add this to the drupal composer project?

rodrigoaguilera:
I submitted a PR with the idea from last comment
https://github.com/drupal-composer/drupal-project/pull/286

I think it will make this one obsolete.

AlexSkrypnyk:

PR #217, opened 2016-11-06
Minor README improvement: embed only valid json

jorissteyn:
Invalid json is displayed on a distracting bright red background. This commit removes the ellipsis from the example json - people will have to understand they have to merge the existing commands in the object.


webflo:
We could wrap the ellipsis in double quotes. "..."

jorissteyn:
Sure, I'll follow up with the addition

webflo:
Thanks!

PR #233, opened 2017-01-12
Ignore web/libraries directory

rwanyoike:
The `composer.json` in the project allows placing components marked as `drupal-library` in the
correct directory, ignore this directory. Related to #215


Schnitzel:
+1 !

PR #239, opened 2017-01-29
also ignore /web/libraries/

Schnitzel:
libraries are also managed by composer: https://github.com/drupal-composer/drupal-project/blob/8.x/composer.json#L63

therefore we can ignore them


Schnitzel:
actually just found #233

PR #245, opened 2017-03-06
Prefix paths in .gitignore with a slash

pfrenssen:
In #178 the settings files were ignored, but this PR predated #210 which prefixed all paths in `.gitignore` with a slash so they do not affect similar paths in subfolders.

This PR again prefixes all paths with slashes so they are limited to the intended files.


webflo:
Thanks!

PR #247, opened 2017-03-09
Add update db after updating modules.

rodrigoaguilera:
I feel is a very nice thing to add to the workflow to ensure that the database is consistent.
I'm thinking of cases when this may be to drastic but I can't come up with any.
I even add a drush config-export after the updb but I guess that really depends on the workflow.

Should I open an issue apart from the Pull request?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,7 +54,8 @@
             "DrupalProject\\composer\\ScriptHandler::createRequiredFiles"
         ],
         "post-update-cmd": [
-            "DrupalProject\\composer\\ScriptHandler::createRequiredFiles"
+            "DrupalProject\\composer\\ScriptHandler::createRequiredFiles",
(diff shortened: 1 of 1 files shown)


rodrigoaguilera:
Oh I see that "post-update-cmd" also happens on composer install.
I will try "post-package-update" but that maybe launches too many drush updb commands

derhasi:
IMHO I do not think this is a good idea in general. That could fit in your use case, but from my perspective composer is there in the first place for managing the code of your project and not the state/database of your drupal installation.

In some cases you would not even have drupal installed at the time you update your dependencies. In others, like often in development, you just want to check the source code before dealing with updates or configs.

I really think code and state should be separated.

webflo:
I agree with @derhasi. Thanks for your suggestion @rodrigoaguilera

PR #252, opened 2017-03-17
Adapt form changes for update notification during site installation

pfrenssen:
In [Issue #2691971: "Update Notifications" #title repeated for drupal 8 installation in configure site page](https://www.drupal.org/node/2691971) the form got reworked that allows to configure whether notifications should be sent about module updates during site installation. This will soon be released with Drupal 8.3.0.

This PR adopts these changes to the installer, so the builds can still run without errors on systems that have no mail transfer agent installed.


pfrenssen:
Woops wrong fork :confused:

PR #255, opened 2017-03-22
Store site configuration outside of the docroot

webflo:
Lets add the <code>$config_directories</code> by default. The installer creates the directory automatically. Its not not required to create it in the script handler itself.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,9 @@
         "cweagans/composer-patches": "~1.0",
         "drupal/core": "~8.0",
         "drush/drush": "~8.0",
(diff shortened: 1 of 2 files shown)


webflo:
<code>webmozart/path-util</code> is already a drush dependency. But i will add it as a root dependency.

webflo:
I realised its possible to get rid of ScriptHandler::getDrupalRoot

pfrenssen:
Looking good to me. Code looks good and I also tried it out and it works beautifully!

webflo:
Thanks! For the review, rebased and committed to 8.x

PR #259, opened 2017-03-28
Fix .gitignore syntax to actually ignore the lock file

mikran:


webflo:
I think its better to remove this line altogether.

madalinignisca:
Agree, that should not be there at all.

@mikran would you update and I'll accept the pull request?

madalinignisca (approved):

PR #264, opened 2017-04-13
Use drupal-core-strict.

greg-1-anderson:
With the [current twig regression](https://www.drupal.org/node/2869528), Composer-managed Drupal sites will get a too-recent version of Twig, and experience problems.

Should we perhaps prevent this by using the `webflo/drupal-core-strict` project? Seems to work well and solve the problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ install:
   - composer --verbose install
 
 script:
-  - if [[ $RELEASE = dev ]]; then composer --verbose require --no-update drupal/core:8.4.x-dev; fi;
+  - if [[ $RELEASE = dev ]]; then composer --verbose require --no-update drupal/core:8.4.x-dev webflo/drupal-core-strict2:8.4.x-dev; fi;
   - if [[ $RELEASE = dev ]]; then composer --verbose update; fi;
   - cd $TRAVIS_BUILD_DIR/web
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
Haven't analyzed why dependency resolution doesn't succeed. Worked in https://github.com/pantheon-systems/example-drops-8-composer/pull/75.

webflo:
Related to https://github.com/composer/composer/issues/6131 and https://github.com/composer/composer/pull/6139

greg-1-anderson:
That's odd.  As I mention above, pantheon-systems/example-drops-8-composer#75 is working, and that project is a derivative of this project. I don't know what the difference is.

webflo:
I build a new package, which has drupal/core as a dependency. I hope it helps to resolve the dependencies in a reliable way.

greg-1-anderson:
This looks good to me. If drupal-core-strict requires drupal-core, then should we advise that it be used in place of, instead of in addition to, drupal-core in a project's composer.json?

I don't know if it makes any difference to Composer, but it seems logical.

webflo:
I think having it in addition is fine, because its an optional package. The DX is not ideal, its makes updates via composer difficult due the related composer issues.

webflo:
Lets keep it optional. It will hurt DX a lot. ``webflo/drupal-core-require-dev`` causes enough confusion for now.

PR #265, opened 2017-04-19
default.services.yml should not be copied to services.yml

mikran:
I just deleted services.yml from my project as I am not overriding anything from core, source on that: https://www.drupal.org/node/2547447.

Running composer install afterwards re-created the services.yml and for reasons mentioned in that core issue this should not happen.


webflo:
Awesome, i did not know that :)

webflo:
👍

PR #268, opened 2017-05-01
Added .htaccess to root to redirect all requests to web directory.

thom8:
This is mainly for safety as it would still be recommended to use `web` as the docroot.

However, in the case where hosting doesn't allow files outside the "docroot" this will redirect all requests to the `web` directory.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -1,0 +1,8 @@
+#
+# Redirect all requests to the web directory.
+#
+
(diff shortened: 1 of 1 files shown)


mathieutu:
Guy, I made the exact same PR here: https://github.com/drupal-composer/drupal-project/pull/294
^^"

PR #270, opened 2017-05-03
Update to Drush 9 for 8.4.x

webflo:


.travis.yml
```diff
@@ -26,7 +32,8 @@ install:
   - composer --verbose install
 
 script:
-  - if [[ $RELEASE = dev ]]; then composer --verbose require --no-update drupal/core:8.4.x-dev; fi;
+  - if [[ $RELEASE = dev ]]; then composer --verbose remove drupal/console; fi;
```
yanniboi: Probably worth adding **--no-update** to the remove drupal/console?

yanniboi:
I think removing the test for php 5.5 is fine. Drupal 8 does technically support >=5.5.9 but there are often bugs in that....

webflo:
The remaining is related to drush site-install on Drupal 8.4.x. Happy if someone whats to tackle this problem.

weitzman:
I'm guessing that the sqlite support in drush9 is lacking. I doubt anyone has exercised it much. Maybe switch to mysql for now?

.travis.yml
webflo: Remove has no such option.

yanniboi: The documentation says it should...?

https://getcomposer.org/doc/03-cli.md#remove

--no-update should remove drupal/console from the composer.json without triggering an update which happens in the next line.

yanniboi:
I would happily look at swapping to mysql for now. But I wont have time for a couple of days. So if someone else wants to have a look in the mean time, then feel free.

webflo:
The bug is not related to sqlite or mysql. The root cause is the changed behaviour of ``drupal_get_profile()`` in 8.4.x

yanniboi:
Thanks @webflo I'm just testing drush 9.0.0-alpha1. I think that will work with dependencies, sqlite, and php 5.5.9 (if we want...)

yanniboi:
Ok, it works. Use this:

`- if [[ $RELEASE = dev ]]; then composer --verbose require --no-update drupal/core:8.4.x-dev drush/drush:~9.0; fi;`

yanniboi:
I've tested the above with mysql (https://travis-ci.org/yanniboi/drupal-project/jobs/229109860) but should also work with sqlite (I assume)

yanniboi:
Also, wrap the last line of the travis file:

`  - ./../vendor/bin/drupal`

in an 

`if [[ $RELEASE = stable ]]`

since we removed drupal console for dev release.

yanniboi:
For reference (https://github.com/yanniboi/drupal-project/pull/1) while this is using mysql (which we don;t need), this is now passing tests.

webflo:
Alright its done, thanks everyone.

yanniboi:
Great work @webflo thanks!

PR #272, opened 2017-05-04
By yanniboi: Updated composer drush version to 9.

yanniboi:
Trying to fix the composer dependency jumble with drupal 8 dev (8.4.x-dev) which seems to now be liking the drush version...

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "drupal-composer/drupal-scaffold": "^2.2",
         "drupal/console": "~1.0",
         "drupal/core": "~8.0",
-        "drush/drush": "~8.0",
+        "drush/drush": "~9.0",
         "webflo/drupal-finder": "^0.2.1",
         "webmozart/path-util": "^2.3"
     },


yanniboi:
hmmm... no that's still failing

yanniboi:
Ok, so I've found the problems... Drupal 8.4.x now uses symfony/yaml v3.2.6. Drush 8 does not support this, but we can upgrade to using an alpha release of drush 9 which will support the new version of yaml.

More of an issue is that drupal console does not support symfony/http-foundation v3.2.6 (also a requirement of drupal 8.4.x), not even the latest dev. So drupal-project will currently now build with drupal console.

What would you prefer? Not support latest dev or not require console?

webflo:
The template is designed to work with 8.3.x atm. I think should remove drupal/console for 8.4.x tests and drop PHP 5.5 because its not compatible with Drush 9. Take a look at the test failures in https://github.com/drupal-composer/drupal-project/pull/270

yanniboi:
I am closing this as a duplicate of #270

PR #276, opened 2017-05-09
Add some configuration to composer.json

benjifisher:
Fix the indentation on the sort-packages line.
Set optimize-autoloader to true.
Set preferred-install to "dist".

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,7 +40,9 @@
     "minimum-stability": "dev",
     "prefer-stable": true,
     "config": {
-      "sort-packages": true
+	"sort-packages": true,
+        "optimize-autoloader": true,
+        "preferred-install": "dist"
     },
     "autoload": {
         "classmap": [


benjifisher:
The `config` section of `composer.json` is often overlooked.  I see you have already added `"sort-packages": true`, and I like that setting a lot.  (When I get a +1 -1 because the item that used to be the last needs a comma at the end, it bugs me.)

I suggest adding

1. `optimize-autoloader: true`
2. `preferred-install: "dist"`

With (1), the autoloader is always optimized.  In my experience, this does not add a noticeable amount of time on Drupal projects, although it does create a couple of large files.  Setting the option avoids see-saw commits adding and removing the autoload files, as in the screenshot in [Update Your Project](https://pantheon.io/docs/guides/github-pull-requests/#update-your-project).

With (2), you usually avoid downloading packages as repositories.  I hate it when I accidentally add a git submodule to my repository.

webflo:
Is preferred-install "dist" much faster than the default "auto"?

fafnirical:
Regarding the `optimize-autoloader` setting, why would one be committing the autoload files? They're in the `/vendor/` directory, which is ignored by git.

benjifisher:
@webflo :  I am not sure, since the [composer docs](https://getcomposer.org/doc/06-config.md#preferred-install) are not explicit on what `auto` does.  I have not tested, but I guess that it has the effect of running [composer install](https://getcomposer.org/doc/03-cli.md#install) with neither `--prefer-source` nor `--prefer-dist`.  If I read *those* docs correctly, then this means downloading a git repo for anything other than a tagged release (or "stable version").

My concern is avoiding git repositories, not the speed of the download.  I am not sure how git submodules work, so I try to avoid them when I can.  If I download something as a git repo and then commit it, then (I think) by default it is added as a submodule.

This is related to the second question ...

@fafnirical, one strategy for deploying is to start with the source repo, then build (`composer install` and build front-end assets) and commit everything to a separate branch, then push that to the remote server.  (See the screenshot I mentioned in the second comment.)  I think that both Pantheon and Platform.sh (less sure) use this method, and maybe also Acquia (CD and "pipelines" ... I have not had a chance to look closely).

Also, if I am developing locally, I prefer to build (by default) closer to the production environment.  The configuration in `composer.json` is not all about what gets committed!

AlexSkrypnyk:

PR #277, opened 2017-05-10
Adding a new create-project-symlinks command

phayes:
When creating a drupal project, it is very common to create a site-specific set of modules, themes, and installation profiles.  

This patch adds a composer command `create-project-symlinks` that looks in the project's root directory for `module`, `theme`, and `profile` directories, and if it finds them, symlinks them into the correct place in the web directory.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -49,6 +49,7 @@
     },
     "scripts": {
         "drupal-scaffold": "DrupalComposer\\DrupalScaffold\\Plugin::scaffold",
+        "create-project-symlinks": "DrupalProject\\composer\\ScriptHandler::createProjectSymlinks",
         "pre-install-cmd": [
             "DrupalProject\\composer\\ScriptHandler::checkComposerVersion"
         ],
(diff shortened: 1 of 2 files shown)


pfrenssen:
I think this is a pretty strange way of approaching it. Drupal-project is already meant to be project specific. There is no reason to not put the custom code in the correct folders from the start, and instead hide them away in some different location. The scaffolding has been provided exactly so you can put everything in the right spot.

If you are running multiple projects inside of one instance of drupal-project and are swapping out the symlinks then it seems to be not being used as intended.

mminklet:
I don't agree, it seems to me like a standard approach to completely ignore the entire drupal project and symlink in custom modules/themes/config. At the moment there is a lot of boilerplate that would be included in the repo if you use the current .gitignore in this project.

For example, why is autoload.php, index.php, robots.txt, update.php not ignored? Arguably there is nothing in the web/ directory that needs to be in the repo. @phayes approach is preferable imo and if you were using the old drush make approach then you would have to do this.

jcnventura:
I agree that autoload.php, index.php, etc should be ignored... Let's fix that and as @pfrenssen says simply put the files in the right place, instead of symlinking them.

AlexSkrypnyk:

PR #286, opened 2017-06-25
Add asset-packagist repository

### LINKED ISSUE #278: Add support for installing libraries via asset-packagist
Let's support adding libraries via asset-packagist.

 If the setup would be done by drupal-project, adding libraries would be simple. We are using this approach for adding module-needed libraries via composer for a while at drunomics now and it works great. It's been adopted by thunder 2.x as well, see https://github.com/BurdaMagazinOrg/thunder-project/pull/5/files.

Once this is done, you can add libraries via a simple command:

`composer require bower-asset/dropzone=~4.2`

More explanation:  https://asset-packagist.org/site/about

rodrigoaguilera:
Be able to download download third party assets with commands like
composer require bower-asset/dropzone=~4.2

closes #278

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,31 +143,42 @@ To prevent this you can add this code to specify the PHP version you want to use
     }
 },
 ```
+
 ### How can I add libraries using composer.json?
 
-It is possible to use libraries with composer thanks to the
-asset-packagist repository (https://asset-packagist.org/).
-
-For example, to use colorbox:
+You can manage front-end asset libraries with Composer thanks to the
+[asset-packagist repository](https://asset-packagist.org/). Composer will detect
+and install new versions of a library that meet the stated constraints. In the
+example below, it will download anything from the `0.4.*` series of colorbox:
 ```
-composer require npm-asset/colorbox:"^0.4"
-
+composer require npm-asset/colorbox:^0.4
 ```
-Composer will detect new versions of the library that meet your constraints.
-In the above example it will download anything from `0.4.*` series of colorbox.
 
-When managing libraries with composer this way, you may not want to add it to
-version control. In that case, add specific directories to the .gitignore file.
+When managing libraries with Composer, it is recommended to add the corresponding
+directories to your `.gitignore` file:
 ```
-# Specific libraries (which we manage with composer)
-docroot/libraries/colorbox
+# Libraries (managed with Composer)
+/docroot/libraries/
 ```
-The path where the library ends up being installed can be controlled too. For example the chose module expects the library at `/libraries/chosen` but with a `composer require npm-asset/chosen-js` the library will have the path `/libraries/chosen-js`.
 
-A library specific override can be added to the installer-paths configuration, otherwise library gets installed with an incorrect name.
-
-Required line: "docroot/libraries/chosen": ["npm-asset/chosen-js"],
-This line must be above the other one to be picked up first.
-
+The installation path of a specific library can be controlled by adding it to the
+`installer-paths` configuration above the general configuration. For example, the
+chosen module expects the library at `/libraries/chosen`, but `composer require
+npm-asset/chosen-js` installs the library into `/libraries/chosen-js`; the
+following override installs it into the expected folder:
+````json
+{
+    "extra": {
+        "installer-paths": {
+            "docroot/libraries/chosen": ["npm-asset/chosen-js"],
+            "docroot/libraries/{$name}": [
+                "type:drupal-library",
+                "type:npm-asset",
+                "type:bower-asset"
+            ]
+        }
+    }
+}
+```
 
 For more details, see https://asset-packagist.org/site/about
--- a/README.md
+++ b/README.md
@@ -166,7 +166,7 @@ The installation path of a specific library can be controlled by adding it to th
 chosen module expects the library at `/libraries/chosen`, but `composer require
 npm-asset/chosen-js` installs the library into `/libraries/chosen-js`; the
 following override installs it into the expected folder:
-````json
+```json
 {
     "extra": {
         "installer-paths": {
--- a/README.md
+++ b/README.md
@@ -162,5 +162,12 @@ version control. In that case, add specific directories to the .gitignore file.
 # Specific libraries (which we manage with composer)
 docroot/libraries/colorbox
 ```
+The path where the library ends up being installed can be controlled too. For example the chose module expects the library at `/libraries/chosen` but with a `composer require npm-asset/chosen-js` the library will have the path `/libraries/chosen-js`.
+
+A library specific override can be added to the installer-paths configuration, otherwise library gets installed with an incorrect name.
+
+Required line: "docroot/libraries/chosen": ["npm-asset/chosen-js"],
+This line must be above the other one to be picked up first.
+
 
 For more details, see https://asset-packagist.org/site/about
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ the [Documentation on drupal.org](https://www.drupal.org/node/2471553).
 First you need to [install composer](https://getcomposer.org/doc/00-intro.md#installation-linux-unix-osx).
 
 > Note: The instructions below refer to the [global composer installation](https://getcomposer.org/doc/00-intro.md#globally).
-You might need to replace `composer` with `php composer.phar` (or similar) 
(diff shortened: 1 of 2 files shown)


fidelix:
This works very well for me. I can recommend.

rodrigoaguilera:
Acquia recently released a blog post explaining this method.
https://lightning.acquia.com/blog/round-your-front-end-javascript-libraries-composer

I hope it becomes a standard workflow.

I will rebase the PR ASAP.

FatherShawn:
I'm also using this method in my projects

hawkeyetwolf:
Adding `component` into the `installer-types` and library path definition would also add support for all the packages in the [components org](https://github.com/components) on Packagist proper.

hawkeyetwolf:
This approach doesn't work for contrib modules right now because they can't safely require asset-packagist dependencies. Per the [composer docs](https://getcomposer.org/doc/05-repositories.md#repository):
> Repositories are only available to the root package and the repositories defined in your dependencies will not be loaded. Read the FAQ entry if you want to learn why.

And this causes two problems:
1. Breaks testbot, since she doesn't have the asset-packagist repository definition in her root project (right now).
1. Prevents requiring the module via Composer by any site which doesn't have the asset-packagist repository defined. We can put big bold words on the d.o project page and in the README, but it still seems problematic.

What do you think? Is there any way around this?

sun:
@derekderaps As repositories are custom by design in Composer, this seems to be the only viable solution – at least for now.

For the time being, this PR resolves https://www.drupal.org/project/drupal/issues/2873160

@rodrigoaguilera Any chance you could rebase it to resolve the conflicts?

weitzman:
In addition to Lightning, Drupal Commerce now ships with this approach - https://www.drupal.org/project/commerce/issues/2916058. +1 from me

heddn:
I was originally a little reluctant to add this to composer. Reason for opposition: my theme needs certain css and fonts. Using bower (npm these days) it is fairly easy to require them. And then when anyone consumes my theme, they have everything packaged for them. In the theme. Just run `yarn install && yarn run gulp` and now I've got my theme setup.

Repositories are set on the base composer.json, so the repository in the theme wouldn't be available. And I'd still need a package.json for gulp, etc. So, at the end of the day, while perhaps useful in some small ways, I'm a little hesitant to think it will solve all dependency management issues on a site.

For the commerce example, that was done on project base. Which means that if someone requires drupal/colorbox, they still need to know to also `composer require npm-asset/colorbox:^0.4`. One doesn't just automatically get the npm asset, or do they?

rodrigoaguilera:
Rebased :)

> For the commerce example, that was done on project base. Which means that if someone requires drupal/colorbox, they still need to know to also composer require npm-asset/colorbox:^0.4. One doesn't just automatically get the npm asset, or do they?

The extra step is still needed.

Maybe in the future contrib projects can start assuming the asset repository is added to your root composer.json just like https://packages.drupal.org/8 

or since the software that manages the  https://asset-packagist.org endpoint is BSD licensed it can be integrated into packages.drupal.org.

rodrigoaguilera:
I do not understand why travis fails. Can someone help with that?

webflo:
@rodrigoaguilera I created a core issue to fix the failing tests. https://www.drupal.org/project/drupal/issues/2960214

composer.json
```diff
@@ -59,9 +64,17 @@
...
+        "installer-types": [
+            "npm-asset",
+            "bower-asset"
+        ],
         "installer-paths": {
             "web/core": ["type:drupal-core"],
-            "web/libraries/{$name}": ["type:drupal-library"],
+            "web/libraries/{$name}": [
+                "type:drupal-library",
+                "vendor:npm-asset",
```
mxr576: I would rather use 

`
"type:bower-asset",
"type:npm-asset"
`
instead. (Just as Lightning and Commerce2 do.)

README.md
```diff
@@ -142,3 +142,24 @@ To prevent this you can add this code to specify the PHP version you want to use
...
 },
 ```
+### How can I add libraries using composer.json?
+
+It is possible to use libraries with composer thanks to the
+asset-packagist repository (https://asset-packagist.org/).
+
+For example, to use colorbox:
+```
+composer require npm-asset/colorbox:"^0.4"
```
mxr576: I'd also mention chosen module and its JS dependency as an example, because if you install the library from npm-asset repo then you have to add a library specific override to the installer-paths, otherwise library gets installed with an incorrect name `chosen-js`.
Required line: ` "docroot/libraries/chosen": ["npm-asset/chosen-js"],`
See:
* https://cgit.drupalcode.org/chosen/tree/README.txt?h=8.x-2.x#n77
* https://github.com/oomphinc/composer-installers-extender

mxr576 (changes requested):

rodrigoaguilera:
I made changes based on @mxr576 suggestions. Please review them.

README.md
```diff
@@ -142,3 +142,42 @@ To prevent this you can add this code to specify the PHP version you want to use
...
+# Libraries (managed with Composer)
+/docroot/libraries/
+```
+
+The installation path of a specific library can be controlled by adding it to the
+`installer-paths` configuration above the general configuration. For example, the
+chosen module expects the library at `/libraries/chosen`, but `composer require
+npm-asset/chosen-js` installs the library into `/libraries/chosen-js`; the
+following override installs it into the expected folder:
+````json
```
sun: Dang, one backtick too much!  Sorry for that.

@rodrigoaguilera Can you quickly fix this in your branch of this PR?

sun: Background info: This causes the closing backticks to no longer match; i.e., all following text is formatted as code.

rodrigoaguilera: Done!

sun (approved):

rodrigoaguilera:
Is it normal to run out of memory for php 5.6?

rodrigoaguilera:
So probably the plugins that we add increase the memory use for php 5.6.

Any ideas on how we can solve it?

From what I can gather right now is either we reduce the memory usage upstream or increase the allowed memory size for tests.

sun:
2 of 3 builds on PHP 5.6 passed though.  Could also have been a hiccup of the build platform.  I do not see a button to re-test; any way to do run it again?

rodrigoaguilera:
I think the history of the PR hides it because of rebases it but it already failed twice and other PRs are not failing.

My only guess is that the composer plugin is raising the memory usage.
Maybe a PR without the plugin can clear the doubts.

Another idea is to make npm-asset and bower-asset "official" types so the plugin is not needed.
I opened an issue for that. 
https://github.com/composer/installers/issues/394

sun:
A completely different question would be whether drupal-composer/drupal-project still needs to support PHP 5.6 in the first place.  In my opinion, attempting to support it is a complete waste of time and resources.  Today, anyone building a new Drupal site with Composer based on this project template will definitely use PHP 7+.

rodrigoaguilera:
I am almost sure the plugins drupal-project depends on do not support php 5.6. 
I feel the dependency for drupal-project on php > 7.0 should be stated more clearly as is creating confusion for some folks.
https://github.com/drupal-composer/drupal-project/issues/403

One of the first steps would be to remove the travis tests for php 5

rodrigoaguilera:
Good to know the memory thing was just a temporary thing.

Can we move this forward? I think it would be great for modules integrating a third party library to assume the repo is in the root composer.json and just declare a dependency.

An example of this situation:
https://www.drupal.org/project/selectize/issues/2949060#comment-12774070

shrop:
I have been working with another developer on this setup in drupal-project and it is really great. Based off the https://lightning.acquia.com/blog/round-your-front-end-javascript-libraries-composer blog post. Would love to see this land here!

kay-o:
looks like this can be merged; anyone aware of remaining tasks? happy to help move this a step closer to being 'standard practice' if so.

rodrigoaguilera:
Are the failures of the CI related to this change?

rodrigoaguilera:
Here is an example of this workflow being adopted by a contrib module.

https://www.drupal.org/project/image_sizes

rodrigoaguilera:
I just found a Drupal issue that has the potential to simplify this PR by just allowing assets to be installed in /vendor and referencing them with a file stream wrapper.

https://www.drupal.org/project/drupal/issues/3002647

AlexSkrypnyk:

PR #292, opened 2017-07-24
Set the default platform version

timmillwood:
Set the default platform version to Drupal's minimum requirement.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,7 +40,10 @@
     "minimum-stability": "dev",
     "prefer-stable": true,
     "config": {
-      "sort-packages": true
+      "sort-packages": true,
+      "platform": {
+        "php": "5.5.9"
+      },
     },
     "autoload": {
         "classmap": [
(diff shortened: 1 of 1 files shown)


jcnventura:
@timmillwood Isn't this the wrong way to do this? If you want to require PHP >= 5.5.9, you need to add it in the require section (see the 7.x composer.json: https://github.com/drupal-composer/drupal-project/blob/7.x/composer.json). Adding it in config seem to be a way to let you fake platform packages in your environment (https://getcomposer.org/doc/06-config.md#platform).

jcnventura:
I'm going to close this however, since Drupal 8's core/composer.json file already requires this, and this project requires drupal/core ^8.0. (http://cgit.drupalcode.org/drupal/tree/core/composer.json)

I'd remove the explicit php requirement in the 7.x branch also, except that Drupal 7 doesn't provide a composer.json file.

PR #294, opened 2017-08-08
Redirect root to /web

mathieutu:
Add a .htaccess in the root folder to be redirected if you visit directly this folder.

Thx for your work.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -1,2 +1,2 @@
 RewriteEngine on
-RewriteRule ^$ /web/
+RewriteRule ^$ /web/ [L,R=301]
--- a/.htaccess
+++ b/.htaccess
@@ -1,0 +1,2 @@
+RewriteEngine on
+RewriteRule ^$ /web/


mathieutu:
Some news about that ?

derhasi:
@mathieutu can you explain why you need this?
This project is intended to only have the web folder being accessible via the web.

mathieutu:
In my case it's for development, I replaced standard drupal sources by yours, but I'm still using the drupal docker container. It permit me to be directly redirected on web without any special config.

As you want, it's just in case, because it can be useful for some people.

worthwhileindustries:
Seems like kind of a unique situation. I'm not sure why you wouldn't just configure the docroot to be web?

webflo:
Its not a sane default for this project imo. https://github.com/wodby/docker4drupal has great docker containers and docroot ist configurable via a environment variable.

mathieutu:
Ok, no problem.
Thx!

PR #297, opened 2017-08-16
Add boilerplate settings files

szeidler:
Adds
* settings.php
* settings.ramsalt.dev.php
* settings.ramsalt.local.php.removeme
* settings.ramsalt.prod
* settings.ramsalt.stage.php

settings.ramsalt.local.php.removeme is used, because settings.ramsalt.local.php is ignored and we don't want to track the file. In addition will the file be renamed on installation.

Let's see what Travis thinks. Any suggestion and changes are welcome.


szeidler:
Oh, that went into the wrong repository.

PR #304, opened 2017-09-04
Add support for Drush 9 and Drupal 8.4

pfrenssen:
Drupal 8.4.x works great with Drush 9 dependency wise. For people with a lot of dependencies it might be easier to resolve to a working set of dependencies if we also allow Drush 9 as an option.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,7 @@ install:
 
 script:
   - if [[ $RELEASE = dev ]]; then composer --verbose remove --no-update drupal/console; fi;
+  - if [[ $RELEASE = dev ]]; then composer --verbose require --no-update drupal/core:8.4.x-dev; fi;
   - if [[ $RELEASE = dev ]]; then composer --verbose update; fi;
   - cd $TRAVIS_BUILD_DIR/web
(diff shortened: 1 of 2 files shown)


webflo:
But it does not work the other way around. This change would install drush 9 for drupal core 8.3.x

webflo:
I works because we have prefer-stable enabled.

dillix:
@webflo now we have stable 8.4 core.

Axolotlke:
In composer.json there is still a refererence to symfony/css-selector 2.x.
This should also be made compatible with the 3.x components required by Drupal 8.4:
`symfony/css-selector (~2.8|~3.0) `

pfrenssen:
OK thanks for the report. I will update the CSS selector.

By the way Drush 8 is now also compatible with Drupal 8.4, so we can declare support for both.

pfrenssen:
We already had support for Drush 8. Updated the CSS selector requirement and also bumped the minimum Drush 9 version to the latest beta. I think this will probably be good to go if this comes back green.

pfrenssen:
Merging this in, 8.4.0 has been out for a few weeks already.

PR #312, opened 2017-10-10
Remove files that are specific to drupal-project after create project

jcnventura:
This patch returns the post-create-project step that was added in PR #7 to the 8.x branch. There's no need to keep the LICENSE and README.txt files. The phpunit.xml.dist and .travis.yml are used only for the project testing and should also be removed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -57,6 +57,9 @@
         ],
         "post-update-cmd": [
             "DrupalProject\\composer\\ScriptHandler::createRequiredFiles"
+        ],
+        "post-create-project-cmd": [
+            "DrupalProject\\composer\\ScriptHandler::removeInternalFiles"
         ]
     },
     "extra": {
--- a/scripts/composer/ScriptHandler.php
+++ b/scripts/composer/ScriptHandler.php
@@ -61,6 +61,27 @@ class ScriptHandler {
   }
 
   /**
+   * Remove project-internal files after create project.
+   */
+  public static function removeInternalFiles(Event $event) {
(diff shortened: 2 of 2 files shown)


fafnirical:
`rm` is not cross-platform -- it doesn't exist on Windows, among possibly other environments. If this does get implemented, it should probably be implemented as a cross-platform PHP script.

pfrenssen:
Good idea. I think we can even replace the existing Travis and PHPUnit manifests with files that will actually test the project itself rather than `drupal-project`. I have these in my fork.

jcnventura:
Well, rm is POSIX.. Most environments that provide a minimum operating system should have rm available.. Of course, Micro$oft being who they are, had to not comply with any standard that they didn't create themselves. But yes, a PHP script would be the way to go.

jcnventura:
@pfrenssen Testing more than the project seems an interesting objective. But I wonder what the design goal should be in this project. I'd rather keep it simple and composer-centered, then to put all the bells and whistles, which require maintenance.

jcnventura:
Updated the code with a PHP-based script instead of using 'rm'..

I've replaced the commit to facilitate a merge.

pfrenssen:
> I'd rather keep it simple and composer-centered, then to put all the bells and whistles, which require maintenance.

Yeah that's the idea. I keep all the bells and whistles in my fork.

But I'm thinking that we can probably move some things like automated testing into the main repo now that we support Drush 9 which depends on robo. Robo is a task runner that can accomplish a lot of stuff.

PR #317, opened 2017-10-13
Ignore unnecessary drupal core files.

jcnventura:
There are several files provided by Drupal 8 that we don't gitignore but don't make any sense as any changes to these probably come from core updates. See the list below of files that would now be ignored.

The only 2 files which might be useful to sometimes add to the repo are .htaccess (for those not running nginx), and robots.txt. I've added commented lines to make it easier for those that want to un-ignore them.

web/.csslintrc
web/.editorconfig
web/.eslintignore
web/.eslintrc.json
web/.gitattributes
web/.htaccess
web/autoload.php
web/index.php
web/robots.txt
web/sites/development.services.yml
web/sites/example.settings.local.php
web/sites/example.sites.php
web/update.php
web/web.config


jcnventura:
@webflo this is not covered by the setup wizard. This is an extension of the ignored files to stop committing some Drupal core files which do not need to be in the repo.

RobLoach (approved):

PR #319, opened 2017-10-17
Make "component" packages install to libraries

RobLoach:
This allows installing packages that are of type "component".

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -65,7 +65,7 @@
     "extra": {
         "installer-paths": {
             "web/core": ["type:drupal-core"],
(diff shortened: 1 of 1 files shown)


hawkeyetwolf:
@RobLoach note that I think this now needs to be updated to add the [Composer Installers Extender](https://github.com/oomphinc/composer-installers-extender) dependency and setting, since we found out the "component" type is not recognized by the main [composer/installers](https://github.com/composer/installers) project.

RobLoach:
Yes, needs updating. Thanks for the reminder.

PR #321, opened 2017-10-20
add more documentation for using composer-patches

redbrickone:
Need to add if you want to want to user composer-patches, you need to first require the plugin.
Added at line 118


jcnventura:
Well, this project includes a base composer.json that already requires that plugin. Other than all the whitespace changes you added, the line explaining to how to require the plugin is redundant.

PR #324, opened 2017-10-26
Make "component" packages install to libraries

RobLoach:
From https://github.com/drupal-composer/drupal-project/pull/319 , allows installation of packages that are of type `component`.

Many of these live over at https://github.com/components .

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,6 +22,7 @@
         "drupal/console": "^1.0.1",
         "drupal/core": "~8.0",
         "drush/drush": "~8.0|^9.0.0-beta7",
+        "oomphinc/composer-installers-extender": "^1.1",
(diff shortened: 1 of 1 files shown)


weitzman:
FYI, asset-packagis is now embraced by both [Lightning](lightning.acquia.com/blog/round-your-front-end-javascript-libraries-composer) and Commerce. Would it make sense to do #286 instead of this PR?

RobLoach:
asset-packagist looks pretty awesome!

hawkeyetwolf:
++asset-packagist. I think _both_ have value in different cases, depending on the package. Would be good to support various options.

RobLoach:
This is great! Going to close this PR.

PR #335, opened 2017-11-29
Implemented paranoia mode [New suggested changes]

jkribeiro:
This PR was based on https://github.com/drupal-composer/drupal-project/pull/180

Initially, I know that is not a good thing to have parallel PR for the same change, I'm gonna explain:
I've been running the paranoia mode on some of my projects for a while and since https://github.com/drupal-composer/drupal-project/pull/180 has been without activity for some time, I decided to fork it and created this PR with the changes that I've made. I tried to create a PR from https://github.com/drupal-composer/drupal-project/pull/180 but it is conflicting, so, here is a fresh one from 8.x.

__Main changes made from the original PR__
- __Recreate the web dir on every assets installer process__: The web directory is disposable, I think its needs to be deleted and recreated from scratch on every installation to remove the old assets and install the new ones.
- __Added new config for Public files folder location__: A new *optional* config was added to be able to set a custom symlink target for the the public files at the web dir. This could help with custom locations on different servers. The new optional config is `drupal-web-dir-public-files`.

I'm running the paranoia mode on Acquia Cloud servers, to be able to customize this for Acquia structure, I've created https://github.com/jkribeiro/drupal-project-acquia-paranoia-mode for additional installation/instruction.


webflo:
Thanks for working on this. I will review it next week. I would like to keep the feature optional (opt-in) because i don't won't to further diverge from a default Drupal setup. It makes it harder for new people to understand all the bits and pieces.

jkribeiro:
I agree, this should be optional. 

@webflo FYI I'm working on an experimental Composer plugin based on this implementation. I think with an separated composer plugin the users can easily install the paranoia mode. Another advantage for having a separated repo for this is that we can isolate the issues, releases etc.

I'll send the repo to you review. If you approve, we can move the plugin to `drupal-composer` org.
Thanks.

jkribeiro:
@webflo as mentioned, I've created a Composer plugin based on this PR.

https://github.com/jkribeiro/drupal-composer-paranoia

Please, let me know what do you think.

jkribeiro:
Hi @webflo , I think we can close this PR because as you said, we don't want to force the paranoia mode to the projects, this should be optional.

We have been using this [plugin](https://github.com/jkribeiro/drupal-composer-paranoia) for about 4 months on 50 sites. Different users are using this and so far no problem has been reported.

What do you think about the idea of moving this plugin to `drupal-composer` GitHub organization as an alpha release?

This plugin/paranoia was mentioned on a DrupalCon session https://events.drupal.org/nashville2018/sessions/drupal-core-auto-update-architecture

webflo:
@jkribeiro Great news. Ok lets move the plugin into the drupal-composer namespace.

webflo:
New Project https://github.com/drupal-composer/drupal-paranoia

PR #343, opened 2018-01-15
Modify the default settings so that if patches cannot be applied the composer install step fails.

vermario:
Hi! I propose this small change:

As found on the readme of the "composer-patches" package, the default behaviour is that a patch cannot be applied the build is allowed to continue.
In my opinion that is not a useful behaviour to have as the default because unpatched modules and core might end up in production, so I propose to set this setting to "TRUE" so that if patching fails the build is not allowed to go through.
People would still be able to remove that to get the old behaviour if desired.

More info on the readme of the composer-patches package at: https://github.com/cweagans/composer-patches#error-handling

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -70,6 +70,7 @@
             "web/profiles/contrib/{$name}": ["type:drupal-profile"],
             "web/themes/contrib/{$name}": ["type:drupal-theme"],
             "drush/contrib/{$name}": ["type:drupal-drush"]
-        }
+        },
+        "composer-exit-on-patch-failure": true
     }
 }


pfrenssen:
This would be very nice to have. If we set this option this will make sure that people that are not aware of this option will not accidentally generate invalid builds if a patch cannot be downloaded for some reason.

In the next major version of composer-patches this option will be set by default. Ref. https://github.com/cweagans/composer-patches/issues/209

pfrenssen:
@vermario is it possible for you to update the PR and solve the merge conflict? The failure that happened seems to be unrelated.

AlexSkrypnyk:
This was added to 10.x
https://github.com/drupal-composer/drupal-project/blob/10.x/composer.json#L82

PR #348, opened 2018-01-28
Replace require-dev with webflo/drupal-core-require-dev

webflo:


wundo:
@webflo why did you do that?

I could not find the reasoning for this change.

This change hides the require-dev dependencies inside an "empty" project, and makes it hard to override the dependencies on the fly. It broke mine (and probably others) continuous integration.

webflo:
@wundo I am sorry about that, but this project its a starterkit. I think you should not depend on it directly. Use it as a base, make it your own.

Back to this specific change. The development dependencies got outdated over time. It was my goal to provide it as and standalone packages. I hope this makes it easier to update your project later on.

PR #350, opened 2018-02-09
Update /drush dir for Drush9

weitzman:
Shows how the benefits of a policyfile and also illustrates where to store site aliases and druh config.


drush/policy.drush.inc
```diff
@@ -1,38 +0,0 @@
...
- * Customize this file as desired. See https://github.com/drush-ops/drush/blob/8.x/examples/policy.drush.inc for documentation.
- */
-
-/**
- * Implements drush_hook_COMMAND_validate().
- *
- * Encourage folks to use `composer` instead of Drush pm commands
- */
-function drush_policy_pm_updatecode_validate() {
-  return _deny_message();
```
FatherShawn, line 13: This functionality is not present in the replacement `Drush\Commands\PolicyCommands` - should we not continue to prevent this?

weitzman: There is no pm-updatecode anymore so no need. Of course people still use Drush8 but thsoe folks are generally not starting brand new projects. This project is a starter-kit so we dont have to worry about existing sites.

FatherShawn: Same for `pm-update` and `pm-download` ...

FatherShawn: Ah - I should have checked the repo, I just did a quick check on _drushcommands.com_ !

webflo:
I did some code-styles fixes on the policy file.

drush/Commands/PolicyCommands.php
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  /**
+   * Unauthorized may not execute updates.
+   *
+   * @hook validate updatedb
+   *
+   * @throws \Exception
+   */
+  public function validateUpdateDb(CommandData $commandData) {
```
webflo: Lets remove the validateUpdateDb example. I think its to advanced. We could like to the example for more examples :) sqlSyncValidate and rsyncValidate is good enough.

drush/Sites/self.site.yml
```diff
@@ -0,0 +1,14 @@
+# Edit or remove this file as needed.
+# Docs at https://github.com/drush-ops/drush/blob/master/examples/example.site.yml
+
+#live:
```
webflo: Lets use prod here. It should match with our default policy file

webflo: drush/**Sites**/self.site.yml should be **sites**?. I looked through the drush code and should not find a instance with ``Sites``.

webflo (changes requested):

weitzman:
Implemented review feedback.

webflo:
@weitzman Thanks! Sorry for the delay.

PR #351, opened 2018-02-09
Honor .env files and suggest usage

weitzman:


webflo:
Thanks @weitzman this is a great addition especially the DRUSH_OPTIONS_URI trick. 👍

sharique:
This is really good feature, can we have some documentation, like how we can use these files for deployment?

weitzman:
Deployment is very site specific and I cant think of any generic docs we can add for it. Do you have anything in mind?

sharique:
Let's say I've created 1 .env file for each environment (like .env.local and .env.dev etc..), 
So what other steps I need to perform before running "composer install",  so that based on environment I get updated settings.php file ready to use. Or what additional parameter I need to pass to "composer install " so that it can generate appropriate settings.php.

weitzman:
`composer install` only downloads your codebase. It doesn't install drupal. Thats a one time step, unlike composer install/update.

You edit your settings.php by hand to add calls to getenv(). See this PR for examples.

PR #353, opened 2018-02-15
Suggest to install Drupal Console

### LINKED ISSUE #352: Move drupal/console from a required dependency to a dev dependency
Drupal Console has some nice functionality but it is not strictly required for drupal-project to function.

I would propose to move it from the `require` section to the `suggest` section. This means that when a Drupal project is created or updated, Composer will output a suggestion to install Drupal Console, and developers can then decide whether to use it or not depending on the needs of their project.

pfrenssen:
Fixes #352.


FatherShawn:
I don't think project itself uses/requires drush either.  If we move this should we not be consistent and suggest Drush as well?  That said **I can't picture setting up a drupal site as CLI administered via Composer without including the other CLI tools for Drupal.**

gnuget:
I agree with @FatherShawn, maybe we can include a fix to #357 on this PR?

pfrenssen:
Good suggestion, Drush is only used in the test suite, so it can be moved to `require-dev`.

weitzman:
require-dev is a poor practice. Most prod sites need a local drush for
routine maintenance.

jackbravo:
I agree with moshe that drush is often used in production environments, although for the purist that is also wrong, but we are talking of things people use, not about some gold standard. For the same reason, moving drupal console to suggested seems backwards to me, maybe, just maybe require-dev would be fine instead of require, and that is just while drush 8 is still being used. I'm not sure right now which one of drush 9 or drupal console I would use more routinely in production, probably console, so I would leave both in require.

pfrenssen:
In the issue there were a lot of people commenting that they are interested in having it available for development by default. Let's move it to `require-dev`. I agree that Drush is probably best kept in `require` since most Drupal sites need it in production.

webflo:
I will keep it as is for now. This could be a feature of the setup wizard. https://github.com/drupal-composer/drupal-project/pull/384

PR #371, opened 2018-03-27
Remove extra whitespace

zaporylie:
This PR removes extra white space so when you copy the code example you won't get coding standard errors.

Nothing special here :)


webflo:
Thx

PR #376, opened 2018-03-28
Updated argument name

morenstrat:
The 'destination' arguments for the sql:sync and core:rsync are now called 'target'.

### DIFF
--- a/drush/Commands/PolicyCommands.php
+++ b/drush/Commands/PolicyCommands.php
@@ -19,7 +19,7 @@ class PolicyCommands extends DrushCommands {
    * @throws \Exception
    */
   public function sqlSyncValidate(CommandData $commandData) {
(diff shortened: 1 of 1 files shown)


morenstrat:
Note: currently the commands don't seem to be picked up. See #377

weitzman:
Are you using Drush9? The argument names match Drush 9, which is what this starter kit installs by default.

morenstrat:
Yes, I'm using Drush 9.2.1 - With argument name `destination` I get ` [error]  The "destination" argument does not exist. `

I moved `drush/Commands/PolicyCommands.php` to `vendor/drush/drush/src/Commands` because otherwise the validation hooks don't get picked up (see #377)

weitzman:
Oh you are correct. I was confused.

Issue #379, opened 2018-04-05
Remove phpdotenv or move it to require-dev


q0rban:
For performance reasons, it's not a good idea to use phpdotenv in production. And especially if the idea of not using it in production is "just don't have a .env, but use phpdotenv to search for one anyway." This choice seems like a choice that an individual project should make, not one that drupal-project should make for everyone using it as a starterkit. And if you do want to use it, it should IMO be a require-dev package, not something that ever makes it to production.

weitzman:
The point of a starter kit IMO is that you show people good practices and you *expect* them to customize it. drupal-project is not forcing any choices on anyone. This is a recurring misconception or disagreement.

q0rban:
The way it's added, it's quite tricky to remove phpdotenv from this project. It's not as simple as `composer remove vlucas/phpdotenv`.

weitzman:
It can be done by commenting out https://github.com/drupal-composer/drupal-project/blob/8.x/composer.json#L44. I agree that we could document this better.

q0rban:
It's actually not as simple as just that either. For the record, in case anyone is reading this that wants to remove it, here are the steps:

1. `composer remove vlucas/phpdotenv`
2. `rm load.environment.php .env.example`
3. Remove `"files": ["load.environment.php"]` from composer.json
4. Run `composer dump-autoload` to regenerate the autoloader.
5. Run `composer update --lock` to update the lock hash, since composer.json was manually edited.

webflo:
@q0rban Thanks for the pull-request. I will think about it, especially because of the long discussions file_exists we had in Drupal core.

There is a shorter version for removing it.

1. `rm load.environment.php .env.example`
2. Remove `"files": ["load.environment.php"]` from composer.json
3. `composer remove vlucas/phpdotenv` (Removes the packages, updates autoloader and lock hash)

q0rban:
Nice, that's much better @webflo! I do think that is unlikely to be people's first thought when removing it, unless they are very familiar with how Composer works.

fidelix:
@webflo your 3 steps don't work. I get a fatal error on step 3:

```
 PHP Warning:  require(/zzz/vendor/composer/../../load.environment.php): failed to open stream: No such file or directory in /zzz/vendor/co  
  mposer/autoload_real.php on line 66 
```

By the way, I support removing this package from the project, or at least moving it to require-dev. The way it is now is just too opinionated.

kevinquillen:
[...]

PR #384, opened 2018-04-13
WIP: Add setup wizard?

webflo:
I wrote a setup class, which provides some configuration options prior project initialization and deletes itself afterwards. This could be interesting for the composer initiative, or evaluator experience. We could add more features to this template without enforcing every feature. 

The code is not pretty, but it does the job.

Current features:

- Change drupal root? web => docroot
- Remove drupal-console / drush / dotenv
- Remove setup wizard itself
- Remove default .travis config and phpunit

Possible features:

- Configure dev env? Lando / ddev / etc.
- Add webflo/drupal-core-strict?

Command for evaluation:

```
composer create-project --repository '{"type": "git", "url": "git@github.com:webflo/drupal-project.git"}' drupal-composer/drupal-project my-project dev-setup -vvv
```

Related issues:

- https://github.com/drupal-composer/drupal-project/issues/379
- https://github.com/drupal-composer/drupal-project/issues/352
- https://github.com/drupal-composer/drupal-project/pull/317
- https://github.com/drupal-composer/drupal-project/pull/312

### DIFF
--- a/scripts/composer/Setup.php
+++ b/scripts/composer/Setup.php
@@ -1,0 +1,69 @@
+<?php
+
+namespace DrupalProject\composer;
+
+use Composer\Json\JsonFile;
+use Composer\Script\Event;
+use Symfony\Component\Filesystem\Filesystem;
+
+class Setup {
+
+  public static function setup(Event $event) {
+    $jsonFile = new JsonFile($event->getComposer()
+      ->getConfig()
+      ->getConfigSource()
+      ->getName());
+    $config = $jsonFile->read();
+    $fs = new Filesystem();
+    $drupalRoot = $event->getIO()->ask('<info>Customize Drupal root path?</info> [<comment>web</comment>]? ', 'web');
+    if ($drupalRoot !== 'web') {
+      $drupalRoot = rtrim($drupalRoot, '/');
+      $gitIgnore = file_get_contents('.gitignore');
+      $gitIgnore = preg_replace('/' . preg_quote('web/', '/') . '/', $drupalRoot . '/', $gitIgnore);
+      if (isset($config['extra']['installer-paths'])) {
+        $installer_paths = [];
+        foreach ($config['extra']['installer-paths'] as $path => $spec) {
+          $newPath = preg_replace('/' . preg_quote('web/', '/') . '/', $drupalRoot . '/', $path);
+          $installer_paths[$newPath] = $spec;
+        }
+        $config['extra']['installer-paths'] = $installer_paths;
+      }
+      $fs->dumpFile('.gitignore', $gitIgnore);
+    }
+    if ($event->getIO()->askConfirmation('<info>Remove dotenv?</info> [<comment>y,N</comment>]? ', TRUE)) {
+      $fs->remove(['.env.example', 'load.environment.php']);
+      if (!empty($config['autoload']['files'])) {
+        $config['autoload']['files'] = array_diff($config['autoload']['files'], ['load.environment.php']);
+        if (empty($config['autoload']['files'])) {
+          unset($config['autoload']['files']);
+        }
+      }
+      unset($config['require']['vlucas/phpdotenv']);
+      unset($config['require-dev']['vlucas/phpdotenv']);
+    }
+    if ($event->getIO()->askConfirmation('<info>Remove drush?</info> [<comment>y,N</comment>]? ', FALSE)) {
+      unset($config['require']['drush/drush']);
+      unset($config['require-dev']['drush/drush']);
+    }
+    if ($event->getIO()->askConfirmation('<info>Remove drupal-console?</info> [<comment>y,N</comment>]? ', FALSE)) {
+      unset($config['require']['drupal/console']);
+      unset($config['require-dev']['drupal/console']);
+    }
+    if ($event->getIO()->askConfirmation('<info>Remove the installer and other scaffold files?</info> [<comment>y,N</comment>]? ', FALSE)) {
+      unset($config['scripts']['setup']);
+      $config['autoload']['classmap'] = array_diff($config['autoload']['classmap'], ['scripts/composer/Setup.php']);
+      $config['scripts']['post-root-package-install'] = array_diff($config['scripts']['post-root-package-install'], ['@setup']);
+      if (empty($config['scripts']['post-root-package-install'])) {
+        unset($config['scripts']['post-root-package-install']);
+      }
+      $fs->remove([
+        '.travis.yml',
+        'phpunit.xml.dist',
+        'scripts/composer/Setup.php',
+      ]);
+    }
+    $jsonFile->write($config);
+    return TRUE;
+  }
+
+}
--- a/composer.json
+++ b/composer.json
@@ -39,12 +39,14 @@
     },
     "autoload": {
         "classmap": [
-            "scripts/composer/ScriptHandler.php"
+            "scripts/composer/ScriptHandler.php",
+            "scripts/composer/Setup.php"
         ],
         "files": ["load.environment.php"]
     },
     "scripts": {
         "drupal-scaffold": "DrupalComposer\\DrupalScaffold\\Plugin::scaffold",
+        "setup": "DrupalProject\\composer\\Setup::setup",
         "pre-install-cmd": [
             "DrupalProject\\composer\\ScriptHandler::checkComposerVersion"
         ],
@@ -56,6 +58,9 @@
         ],
         "post-update-cmd": [
             "DrupalProject\\composer\\ScriptHandler::createRequiredFiles"
+        ],
+        "post-root-package-install": [
+            "@setup"
         ]
     },
     "extra": {


jcnventura:
Would make sense to also remove ~~LICENSE and~~ README.md.

~~Sites that are not distributed may not need to be GPL-licensed.~~

pfrenssen:
> Sites that are not distributed may not need to be GPL-licensed.

They still need to be GPL-licensed even if you don't intend to distribute them to a client.

jcnventura:
@pfrenssen Yes, you're right. I got confused. Of course the site is GPL-licensed even in that case. What happens when you don't distribute the site is that you're not forced to provide the site user with a copy of it's source code.

Still, the drupal-project README.md should be removed from the created project.

scripts/composer/Setup.php
```diff
@@ -0,0 +1,69 @@
...
+      $gitIgnore = preg_replace('/' . preg_quote('web/', '/') . '/', $drupalRoot . '/', $gitIgnore);
+      if (isset($config['extra']['installer-paths'])) {
+        $installer_paths = [];
+        foreach ($config['extra']['installer-paths'] as $path => $spec) {
+          $newPath = preg_replace('/' . preg_quote('web/', '/') . '/', $drupalRoot . '/', $path);
+          $installer_paths[$newPath] = $spec;
+        }
+        $config['extra']['installer-paths'] = $installer_paths;
+      }
+      $fs->dumpFile('.gitignore', $gitIgnore);
```
jcnventura, line 31: Move this before line 23. The logical flow would be to save the new gitignore just after the preg_replace on the gitIgnore lines, and not after the installer_paths block.

scripts/composer/Setup.php
```diff
@@ -0,0 +1,69 @@
...
+    }
+    if ($event->getIO()->askConfirmation('<info>Remove dotenv?</info> [<comment>y,N</comment>]? ', TRUE)) {
```
jcnventura, line 33: Since default is true, it should be 'Y,n'.

scripts/composer/Setup.php
```diff
@@ -0,0 +1,69 @@
...
+      if (!empty($config['autoload']['files'])) {
+        $config['autoload']['files'] = array_diff($config['autoload']['files'], ['load.environment.php']);
+        if (empty($config['autoload']['files'])) {
+          unset($config['autoload']['files']);
+        }
+      }
+      unset($config['require']['vlucas/phpdotenv']);
+      unset($config['require-dev']['vlucas/phpdotenv']);
+    }
+    if ($event->getIO()->askConfirmation('<info>Remove drush?</info> [<comment>y,N</comment>]? ', FALSE)) {
```
jcnventura, line 44: In this case, also remove the drush/Commands/PolicyCommands.php, drush/README.md, drush/drush.yml and drush/sites/self.site.yml files (same as other scaffold files below).

scripts/composer/Setup.php
jcnventura: Also, remove the .env.example and load.environment.php files (same as other scaffold files below).

scripts/composer/Setup.php
```diff
@@ -0,0 +1,69 @@
...
+      $config['autoload']['classmap'] = array_diff($config['autoload']['classmap'], ['scripts/composer/Setup.php']);
+      $config['scripts']['post-root-package-install'] = array_diff($config['scripts']['post-root-package-install'], ['@setup']);
+      if (empty($config['scripts']['post-root-package-install'])) {
+        unset($config['scripts']['post-root-package-install']);
+      }
+      $fs->remove([
+        '.travis.yml',
+        'phpunit.xml.dist',
+        'scripts/composer/Setup.php',
+      ]);
```
jcnventura, line 63: Remove also the drupal-composer README.md file.

scripts/composer/Setup.php
```diff
@@ -0,0 +1,69 @@
...
+    }
+    if ($event->getIO()->askConfirmation('<info>Remove drush?</info> [<comment>y,N</comment>]? ', FALSE)) {
+      unset($config['require']['drush/drush']);
+      unset($config['require-dev']['drush/drush']);
+    }
+    if ($event->getIO()->askConfirmation('<info>Remove drupal-console?</info> [<comment>y,N</comment>]? ', FALSE)) {
+      unset($config['require']['drupal/console']);
+      unset($config['require-dev']['drupal/console']);
+    }
+    if ($event->getIO()->askConfirmation('<info>Remove the installer and other scaffold files?</info> [<comment>y,N</comment>]? ', FALSE)) {
```
jcnventura, line 52: Project would be more user friendly if this option was TRUE by default. In which case, make the comment text 'Y,n'.

scripts/composer/Setup.php
```diff
@@ -0,0 +1,69 @@
...
+    if ($event->getIO()->askConfirmation('<info>Remove dotenv?</info> [<comment>y,N</comment>]? ', TRUE)) {
+      $fs->remove(['.env.example', 'load.environment.php']);
+      if (!empty($config['autoload']['files'])) {
+        $config['autoload']['files'] = array_diff($config['autoload']['files'], ['load.environment.php']);
+        if (empty($config['autoload']['files'])) {
+          unset($config['autoload']['files']);
+        }
+      }
+      unset($config['require']['vlucas/phpdotenv']);
+      unset($config['require-dev']['vlucas/phpdotenv']);
```
jcnventura, line 42: There is no phpdotenv in the require-dev section of composer.json.

pfrenssen:
This works great but it throws an error when it starts installing dependencies because the altered `composer.json` file is not reloaded and it still tries to load the `Setup.php` file which has in the meantime been deleted.

I have been looking through the code if it is possible to trigger a reload of the `composer.json` file (it is tracked in `Config::$configSource`) but I couldn't find any code that allows to do this.

In my own project I solved it by implementing `post-create-project-cmd` hook and delaying the deletion of the file until this hook fires.

Here's the patch I used in my project to fix this, it might help you:

```
diff --git a/composer.json b/composer.json
index 969d52c..38670cb 100644
--- a/composer.json
+++ b/composer.json
@@ -31,9 +31,13 @@
         "post-root-package-install": [
             "@setup"
         ],
+        "post-create-project-cmd": [
+            "@cleanup"
+        ],
         "post-update-cmd": [
             "./vendor/bin/run drupal:site-setup"
         ],
+        "cleanup": "DrupalSiteTemplate\\composer\\Setup::cleanup",
         "setup": "DrupalSiteTemplate\\composer\\Setup::setup"
     },
     "extra": {
diff --git a/scripts/composer/Setup.php b/scripts/composer/Setup.php
index e59d29d..30592ab 100644
--- a/scripts/composer/Setup.php
+++ b/scripts/composer/Setup.php
@@ -75,25 +75,31 @@ class Setup {
       file_put_contents($filename, $file);
     }
 
-    // Remove the setup wizard.
-    unlink('scripts/composer/Setup.php');
-
+    // Remove the configuration related to the setup wizard.
+    unset($config['scripts']['cleanup']);
     unset($config['scripts']['setup']);
     $config['autoload']['classmap'] = array_diff($config['autoload']['classmap'], ['scripts/composer/Setup.php']);
     if (empty($config['autoload']['classmap'])) {
       unset($config['autoload']['classmap']);
     }
+    $config['scripts']['post-create-project-cmd'] = array_diff($config['scripts']['post-create-project-cmd'], ['@cleanup']);
+    if (empty($config['scripts']['post-create-project-cmd'])) {
+      unset($config['scripts']['post-create-project-cmd']);
+    }
     $config['scripts']['post-root-package-install'] = array_diff($config['scripts']['post-root-package-install'], ['@setup']);
     if (empty($config['scripts']['post-root-package-install'])) {
       unset($config['scripts']['post-root-package-install']);
     }
 
     $composer_json->write($config);
 
     return TRUE;
   }
+
+  /**
+   * Removes the setup wizard.
+   */
+  public static function cleanup(): void {
+    unlink('scripts/composer/Setup.php');
+  }
 }
```

FatherShawn:
I added a lot of wizard-like behavior in [digitalpulp/ballast](https://github.com/digitalpulp/ballast) which is downstream to this project. `post-create-command-project-cmd` is a great event - I should have looked at the event list.  I'll move some of my tasks to more appropriate events- thanks for the tip @pfrenssen !

pfrenssen:
I'm also having another problem related to modifying the `composer.json` file: the autoloader is generated using the original file, so I need to execute `$ composer dump-autoload` after running the wizard. I have no time unfortunately right now to look into a structural solution for this.

We are also working on a downstream variant of this, it might serve as inspiration: https://github.com/openeuropa/drupal-site-template/pull/1

hansfn:
Are we about to recreate drupal-init from [hussainweb/drupal-composer-init](https://github.com/hussainweb/drupal-composer-init). Probably not :-)

RaphTbm:
Hi! 

Is the current whole code in `ScriptHandler.php` should be moved in this initial setup too?

I don't have a full scope of it, but the only necessity for `ScriptHandler.php` to be in the `post-install/update` commands seems to be relative to the "Prepare the settings file for installation" part which is dependent of `drupal/core` and what is retrieved by `drupal/scaffold`. 
But to the end, `drupal/core` and the drupal scaffolded files (as `sites/default/default.settings.php`) are installed during the `create-project` process and the files created/updated in `ScriptHandler.php` seems to be related to a need on the initial install/setup only. 

I supposed that `post-create-project-cmd` was not knew previously (?)

edit: for `ScriptHandler.php`, it's not the whole code of the file, but the whole code of `createRequiredFiles()` function!

Cheers!

Chi-teck:
DCG has got similar wizard recently.
`drush generate project --directory=/path/to/project`

normanlolx:
What's DCG?

Chi-teck:
> What's DCG?

https://github.com/Chi-teck/drupal-code-generator

I think the approach with script handlers in Composer template will always be tricky because of the need to update/remove existing files.

AlexSkrypnyk:

PR #385, opened 2018-04-20
Bump minimal version to 8.5.2

zaporylie:


jcnventura:
Not sure this is necessary. Composer will pull 8.5.2 even if the version currently specified is ~8.5.1. That just means to pull 8.5.1 or newer. It does make sense to specify 8.5.1, as the 8.5.0 security vulnerability is severe enough to warrant that composer should never try to install 8.5.0, even if for some reason it can't contact drupal packagist, and it has an 8.5.0 in cache.

zaporylie:
8.5.2 is a security release, therefore we should ensure no one is using Drupal prior that version, even if its severity is * only * 12/25. Main reason for that change is that if composer.json of project's dependency (contrib or custom) boldly defines its version constraints (skipping ^ or ~) drupal/core won't get updated even ~8.5.1 allows it. If we bump it to ~8.5.2 composer will tell user about the version conflict that must be resolved.

Anyways - it is not that important as we're getting one more security release in two days from now https://www.drupal.org/psa-2018-003

jcnventura:
True, but note that composer.json specified ~8.0 until [November 2017](https://github.com/drupal-composer/drupal-project/commit/89a6e99a1a536a9bbd5bb5af65b92f68855348ab). That was after the 8.0.4, 8.1.3, 8.1.7, 8.1.10, 8.2.3, 8.2.8, 8.3.4 and 8.3.7 security releases.

Composer should just do it's job. And yes, that does mean that if someone somewhere creates a dependency on version 8.5.1 or lower on some contrib module, then composer should fall back on that version. This project should not chase ambulances. Unless of course those ambulances are 25/25 critical.

jcnventura:
Closing PR since we'll have 8.5.3 soon :)

webflo:
https://github.com/drupal-composer/drupal-project/commit/24ab11d2bf57929544944a1c29b72c7a2c942e98

I agree with @zaporylie on this one. We should bump the version in this project to enforce the newest version as a starting point.

PR #394, opened 2018-05-25
Fix installer path for drush commandfiles.

weitzman:
See https://github.com/drush-ops/drush/issues/3566


weitzman:
Test fails look unrelated.

PR #420, opened 2018-09-03
Added "composer-exit-on-patch-failure": true option to composer.json

MPParsley:
By default, failed patches are skipped and a message will be shown.

If a patch cannot be applied (hunk failed, different line endings, etc.) I believe it would be useful to enforce throwing an error and to stop package installation/update immediately.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,6 +58,7 @@
         ]
     },
     "extra": {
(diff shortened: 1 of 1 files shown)


daften (approved):

webflo:
Thanks!

PR #423, opened 2018-09-05
Adding Notes about Composer Version Constraints.

jonpugh:
This explains that to upgrade drupal with the `composer update` command you have to change the `drupal/core` version constraint.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,8 @@ If you use the default version constraint for `drupal/core` used in this project
 
 If you want the `composer update` command to automatically update minor versions, like from `8.5` to `8.6` automatically, set the version of `drupal/core` and `webflo/drupal-core-require-dev` to `~8.5`.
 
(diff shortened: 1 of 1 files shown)


jonpugh:
Maybe we should suggest the wildcard version constraint, it's a little easier to understand I think...

If you use `~8.4` it could be 8.5, 8.6, etc. 
If you use `8.*` it could also be 8.5, 8.6, etc, but you don't see the older version and wonder why.

Seems more intuitive.

hansfn:
I'm fine with either version constraint. However, we might not need all this general info as the default for new install is the caret notation ("^8.6.0") which always all minor updates. In the [Drupal.org docs](https://www.drupal.org/docs/8/update/update-core-via-composer) I just wrote: 

> If you are upgrading from 8.5 to 8.6, you need to replace "~8.5.x" with "^8.6.0" for drupal/core and webflo/drupal-core-require-dev in composer.json

That line can eventually be removed ...

AlexSkrypnyk:

PR #424, opened 2018-09-05
Update composer.json to require latest stable version of Drupal

grasmash:
I'd like to post the following instructions on the /download page on Drupal.org:

    composer create-project drupal-composer/drupal-project:8.x-dev --stability dev --no-interaction

However, this will require drupal-project to provide the latest recommended version of Drupal so that it matches the download button.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ install:
 
 script:
   - if [[ $RELEASE = dev ]]; then composer --verbose remove --no-update drupal/console; fi;
-  - if [[ $RELEASE = dev ]]; then composer --verbose require --no-update drupal/core:8.6.x-dev webflo/drupal-core-require-dev:8.6.x-dev; fi;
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
I agree that this is preferable. `composer update` should track the latest stable Drupal version by default.

hansfn:
Maybe we should documented somewhere that drupal-composer/drupal-project no longer is locked to a specific minor version since we have replaced `"drupal/core": "~x.y.z"` with `"drupal/core": "^x.y.z"` 

More importantly, it should be mentioned in the [update section](https://github.com/drupal-composer/drupal-project#updating-drupal-core) that people with existing projects have to make this change to `drupal/core` and `webflo/drupal-core-require-dev` manually.

jonpugh:
Note I submitted a PR to the README.md related to this... #423

PR #433, opened 2018-10-11
Documentation clarification when using local patch files

marklabrecque:
This change would have really benefitted me and saved me a lot of time today, so I think it might be worth mentioning the in the documentation/README. You can see the fix being referenced here: https://github.com/cweagans/composer-patches/issues/146

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,9 @@ section of composer.json:
     }
 }
 ```
+
+One thing to worth noting with local patch files, it appears the patch will only be detected if it has a .txt file extension.
+
 ### How do I switch from packagist.drupal-composer.org to packages.drupal.org?
 
 Follow the instructions in the [documentation on drupal.org](https://www.drupal.org/docs/develop/using-composer/using-packagesdrupalorg).


leymannx:
> One thing to worth noting with local patch files, it appears the patch will only be detected if it has a .txt file extension.

I guess that's wrong. For Drupal that's always the `*.patch` extension.

This for example is the patches section of one of my projects and it works just fine:

    "extra": {
        "patches": {
            "drupal/menu_block": {
                "menu block active trail": "https://www.drupal.org/files/issues/2018-03-13/menu_block-active_trail-2951298-4-D8.patch"
            },
            "drupal/field_group": {
                "Notice: Undefined index: name": "https://www.drupal.org/files/issues/2018-04-30/2968184-03.patch",
                "Add option to set group label classes for HTML element type": "https://www.drupal.org/files/issues/field_group-group_label_class-2863821-7.patch"
            },
            "drupal/tablefield": {
                "TableField Cellspan sub-module": "https://www.drupal.org/files/issues/2018-07-11/2985110-4-tablefield_cellspan.patch"
            },
            "drupal/eu_cookie_compliance": {
                "Notice: Undefined variable: primary_button_label": "https://www.drupal.org/files/issues/2018-07-13/2985543-3-notice_undefined_index.patch"
            },
            "drupal/svg_image": {
                "Responsive image support": "https://www.drupal.org/files/issues/2018-06-22/svg_image_responsive_support-2981263-2.patch"
            }
        }
    }

milkovsky:
This note makes sens in the readme. I am always doing that :)

jcnventura:
Looking at the original issue in cweagans/composer-patches#146, I think this is more a problem of how that local patch file is created, than a change to the documentation.

bradbulger:
Just to get this explicitly stated, you're saying that if I download a patch file from a URL and try to apply it as a local file, I have to recreate it with --no-prefix first? And that the "add .txt" part is irrelevant?

PR #440, opened 2018-12-10
Added FAQ `How do I add a library?`

ChandeepKhosa:


rodrigoaguilera:
Maybe we should try to push forward some of the issues that are already have the same goal

https://github.com/drupal-composer/drupal-project/pull/215
https://github.com/drupal-composer/drupal-project/pull/286
https://github.com/drupal-composer/drupal-project/issues/278

ChandeepKhosa:
Thanks @rodrigoaguilera but I believe it would be helpful for less experienced devs to see what I've added as I also rewrote the documentation I linked to after mentoring some junior devs at work with this. 
I'd like to suggest that this gets merged in ASAP, and then the other issues you've mentioned could be worked on with more time and perhaps replace the FAQ answer I've added.

Issue #408, opened 2018-07-09
Composer can not install on PHP 5.5.x (tested 5.5.13) - reopened see #403


joshua-graham-adelphi:
not sure how to reopen an issue. see #403

geerlingguy:
PHP 5.5 has not been receiving security support for over a year, and even [5.6 will be end-of-lifed in less than 20 days](http://php.net/supported-versions.php). [Drupal 7 somewhat supports older versions](https://www.drupal.org/docs/7/system-requirements/drupal-7-php-requirements) still, but it's extremely highly-recommended to not use an unsupported/old PHP version if at all possible.

If this project's goal is to help in bootstrapping _new_ Drupal projects, there's even more motivation to use a supported PHP version.

Currently Drupal 8 also supports 5.5 and 5.6, but that lifeline will also be gone once 8.7.0 is released on March 6, 2019... so I wouldn't expect anyone to try doing any work to make any of the dependencies which are not currently backwards-compatible with PHP 5.6 backwards-compatible for just a few months of support for an unsupported PHP version.

normanlolx:
I must apologize upfront for the following heretical question:

Why on earth would you try to install Drupal 8 on a PHP version that reached end-of-life more than two years ago?

> A release that is no longer supported. Users of this release should upgrade as soon as possible, as they may be exposed to unpatched security vulnerabilities.

* [PHP Supported Versions](https://secure.php.net/supported-versions.php)
* [PHP Unsupported Branches](https://secure.php.net/eol.php)

While Drupal 8 still may support an outdated PHP version some dependencies listed in this Composer project may already have dropped it. I guess your best bet is to download the mere Zip on https://www.drupal.org/project/drupal and start your project from there (not recommended) or to upgrade your PHP version (recommended).

joshua-graham-adelphi:
Hi @geerlingguy & @leymannx 

I do concur with your opinion, however, [the project page](https://github.com/drupal-composer/drupal-project) implies that the project should run on PHP 5.5.9 as per minimum drupal. Can we have the page updated with minimum PHP version required for this composer project????

geerlingguy:
@joshua-graham-adelphi - Good point. Note that I'm not the maintainer of this project, just a happy user and contributor. Anyone can update the README via a pull request. I would suggest that section be removed entirely, but if it is kept around, it should definitely be changed.

webflo:
[...]

PR #445, opened 2019-01-07
Add note to advise of PHP 5.6 requirement due to dependencies.

joshua-graham-adelphi:
Add note as project can't be used on PHP 5.5.x

@see #408


webflo:
Thanks for the PR. I've decided to rewrite the sentence. Fixed in 4a664d2.

PR #466, opened 2019-02-18
Travis already displays composer version

szepeviktor:
Lately Travis added self-update which also displays the current version.


webflo:
But the tests explicitly updates composer to a different version (e.g. snapshot channel).

PR #472, opened 2019-03-04
Change preserve-paths to use "sites/all" to protect multisite folders.

jonpugh:
In the drupal 7 composer template, running composer install the first time (or when you delete vender folder) will silently remove any sites/X folders. This will destroy any multisites installed, and prevents the codebase from being used with systems like @aegirproject.

Changing the preserve-paths value to "web/sites" worked great for me.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -79,15 +79,7 @@
             }
         },
         "preserve-paths": [
-            "web/sites/all/drush",
-            "web/sites/all/libraries",
-            "web/sites/all/modules/contrib",
-            "web/sites/all/modules/custom",
-            "web/sites/all/modules/features",
-            "web/sites/all/themes/contrib",
-            "web/sites/all/themes/custom",
-            "web/sites/all/translations",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -79,15 +79,7 @@
...
-            "web/sites/all/drush",
-            "web/sites/all/libraries",
-            "web/sites/all/modules/contrib",
-            "web/sites/all/modules/custom",
-            "web/sites/all/modules/features",
-            "web/sites/all/themes/contrib",
-            "web/sites/all/themes/custom",
-            "web/sites/all/translations",
-            "web/sites/default"
+            "web/sites/all"
```
jcnventura: This will not help at all in protecting the multi-site folders, and will ALWAYS delete the settings.php.

Can you revert this commit, please?

jcnventura (approved):
Sorry. Got the commits mixed up. The second commit is right.

PR #483, opened 2019-04-09
Add lando config

geekygnr:
I thought it would be helpful if the repo had a very basic lando configuration. I would encourage similar done for all common tools for quick dev environment setup.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -25,3 +25,6 @@
 
 # Another common use case is to set Drush's --uri via environment.
 # DRUSH_OPTIONS_URI=http://example.com
+
+# This option will make running drush correctly tie xdebug back to the IDE
+# PHP_IDE_CONFIG="serverName=d8.lndo.site"
--- a/.lando.yml
+++ b/.lando.yml
@@ -1,0 +1,7 @@
+name: d8
+recipe: drupal8
+config:
+  webroot: ./web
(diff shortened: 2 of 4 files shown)


normanlolx:
👍 Nice to have, yes. But environment is totally up to every developer themself and goes beyond the scope of this template I guess.

truls1502:
> 👍 Nice to have, yes. But environment is totally up to every developer themself and goes beyond the scope of this template I guess.

I am totally agreed with you, @leymannx that it should be up to every developer.
However, I believe it would be nice to have some popular recipes for development environments such as [DDEV](https://ddev.readthedocs.io) and [Lando](https://docs.devwithlando.io) to let the new beginner try to spin up the Drupal site in just a few secs. It could save a lot of their time, especially first time so they can play more with Drupal then later they can customise after their need. For the advanced developers so we can create a patch with our own customised on these development environments.

geekygnr:
@truls1502 Thank you. I think you put my thought process down in a far more elegant way than I ever could.

AlexSkrypnyk:
@truls1502 Currently, most of the code in the repo is covered with tests. How would you check that what you are proposing is working in CI? 

I think this project is not a place for development environment configurations.

AlexSkrypnyk:

PR #486, opened 2019-04-17
Add macOS "DS_Store" invisible files to .gitignore

mazleo:


normanlolx:
I think you should better ignore them from your own global ignore file. This Composer template is a starter template you then can adjust to match your needs or match your OS.

jcnventura:
We start with macOS .DS_Store, then eventually all other IDE files get added.

This is a path we should not take.

AlexSkrypnyk:
@mazleo 

These kinds of files should never be added to projects and should be excluded locally.

 please see https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

webflo:
I agree with @leymannx @jcnventura and @alexdesignworks. Please configure your global gitignore :)

Thanks all!

Issue #502, opened 2019-07-26
Add minimum required PHP Extensions to composer.json


cybtachyon:
We currently set minimum PHP version to assist DevOps (and developers writing contrib code).

Could we also add the minimum required PHP extensions to run Drupal? This would greatly assist DevOps who are installing Drupal for the first time somewhere, not to mention would help developers writing contrib code who need to know the baseline extensions available.

Something like:
```json
{
  "ext-curl": ">=7.0",
  "ext-gd": ">=7.0",
  "ext-json": ">=1.4.0",
  "ext-mbstring": ">=7.0",
  "ext-openssl": ">=7.0",
  "ext-pcre": ">=7.0",
  "ext-pdo": ">=7.0",
  "ext-tokenizer": ">=7.0",
  "ext-xml": ">=7.0",
  "ext-zip": ">=1.14.0"
}
```

cybtachyon:
It's worth noting that PHP `5.6` is the minimum version that is technically supported by Drupal version `8.7.0` - and the active stable release is `8.7.5` as of July 17th.

https://www.drupal.org/docs/8/system-requirements/php-requirements

https://www.drupal.org/project/drupal

I'd personally advocate even setting these to PHP `7.1` and the extension default versions for `7.1`.

jcnventura:
Once Drupal 8.8 is released, we can bump the minimum core version and the minimum version of PHP to 7.0.8.

https://www.drupal.org/node/3044409

jcnventura:
As to adding these extensions, I don't understand why these would need to be explicitly state them here, when they are already required by core's composer.json.

https://git.drupalcode.org/project/drupal/blob/8.8.x/core/composer.json

PR #503, opened 2019-07-26
#502: Add php extensions requirements

cybtachyon:
https://github.com/drupal-composer/drupal-project/issues/502


joestewart:
That is a different list of extensions than core uses. I'm curious.

https://git.drupalcode.org/project/drupal/blob/8.8.x/core/composer.json#L7-19 and https://github.com/drupal/core/blob/8.8.x/composer.json#L7-L20

cybtachyon:
Good point! I've updated it to be closer to core composer.json as well as https://www.drupal.org/docs/8/system-requirements/php-requirements

Thanks!

andypost:
Btw it looks better as patch to core itself

AlexSkrypnyk:
There is no need to specify these in the project `composer.json` as they are inherited from core.

PR #504, opened 2019-07-26
Update `composer.json` in order to allow development versions instead to set "minimum-stability"

phansys:
Update `composer.json` in order to whitelist (with [stability flags](https://getcomposer.org/doc/04-schema.md#package-links)) development versions instead to set "minimum-stability".

## The problem with the current setup (using "dev" as "minimum-stability"):
- `my-project` requires `package-A:^2.0`, which has a stable release;
- users perform `composer update`
- `package-A` is resolved to their last available 2 major (stable) version, let's say `2.7.4`;
- `package-A` requires `package-B:^1.0`, which has a stable release;
- `package-B` is resolved to their last available 1 major (stable) version, let's say `1.0.0`;
---
Until here, everything should be fine, since `my-project` knows that all the installed packages are stable. Let's see the following:

- in the PATCH version `2.7.5` the `package-A` decides to change their requirement from `package-B:^1.0` to `package-B:2.0-dev`, which **is not a stable** release;
- users perform `composer update`
- `package-A` is resolved to their last available 2 major (stable) version (`2.7.5`);
- `package-B` is resolved to their last available 2 major (**unstable**) version (2.0-dev);
---
At this point, the users which own the responsibility under `my-project` were installed silently an unstable version of a package which can potentially break some existing feature or behavior in their project.
With the proposed changes, if the users want to use the `2.7.5` version from `package-A`, they will be restricted to install the unstable version from package `package-B`, at least if they decide under their own risk and responsibility to trust on that package, whitelisting it until a stable release will be available from what is required from `package-A`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ for your setup.
 After that you can create the project:
 
 ```
-composer create-project drupal-composer/drupal-project:8.x-dev some-dir --no-interaction
+composer create-project drupal-composer/drupal-project:^8.0@dev some-dir --no-interaction
 ```
 
 With `composer require ...` you can download new dependencies to your
@@ -28,7 +28,7 @@ installation.
 
 ```
 cd some-dir
-composer require drupal/devel:~1.0
+composer require drupal/devel:^1.0
 ```
 
 The `composer create-project` command passes ownership of all files to the
--- a/composer.json
+++ b/composer.json
@@ -16,26 +16,28 @@
         }
     ],
     "require": {
-        "php": ">=7.0.8",
+        "php": "^7.0.8",
         "composer/installers": "^1.2",
         "cweagans/composer-patches": "^1.6.5",
         "drupal-composer/drupal-scaffold": "^2.5",
         "drupal/console": "^1.0.2",
-        "drupal/core": "^8.7.0",
-        "drush/drush": "^9.0.0",
+        "drupal/core": "^8.7",
+        "drush/drush": "^9",
         "vlucas/phpdotenv": "^2.4",
-        "webflo/drupal-finder": "^1.0.0",
+        "webflo/drupal-finder": "^1",
         "webmozart/path-util": "^2.3",
-        "zaporylie/composer-drupal-optimizations": "^1.0"
+        "zaporylie/composer-drupal-optimizations": "^1"
     },
     "require-dev": {
-        "webflo/drupal-core-require-dev": "^8.7.0"
+        "behat/mink": "^1.7@dev",
+        "behat/mink-selenium2-driver": "^1.3@dev",
+        "friendsofphp/php-cs-fixer": "^2.15",
+        "phpunit/phpunit": "^6.5",
+        "webflo/drupal-core-require-dev": "^8.7"
     },
     "conflict": {
         "drupal/drupal": "*"
     },
-    "minimum-stability": "dev",
-    "prefer-stable": true,
     "config": {
         "sort-packages": true
     },


composer.json
```diff
@@ -16,26 +16,28 @@
...
-        "zaporylie/composer-drupal-optimizations": "^1.0"
+        "zaporylie/composer-drupal-optimizations": "^1"
     },
     "require-dev": {
-        "webflo/drupal-core-require-dev": "^8.7.0"
+        "behat/mink": "^1.7@dev",
+        "behat/mink-selenium2-driver": "^1.3@dev",
+        "friendsofphp/php-cs-fixer": "^2.15",
+        "phpunit/phpunit": "^6.5",
+        "webflo/drupal-core-require-dev": "^8.7"
```
jcnventura, line 36: All the above are what webflo/drupal-core-require-dev already does.. Please don't do this.

README.md
```diff
@@ -14,25 +14,25 @@ the [Documentation on drupal.org](https://www.drupal.org/node/2471553).
...
 > Note: The instructions below refer to the [global composer installation](https://getcomposer.org/doc/00-intro.md#globally).
-You might need to replace `composer` with `php composer.phar` (or similar) 
+You might need to replace `composer` with `php composer.phar` (or similar)
 for your setup.
 
 After that you can create the project:
 
 ```
-composer create-project drupal-composer/drupal-project:8.x-dev some-dir --no-interaction
+composer create-project drupal-composer/drupal-project:^8.0@dev some-dir --no-interaction
```
jcnventura, line 23: Please revert. It's correct as it is. There is no 8.0 of this project.. There's only 8.x.

README.md
```diff
@@ -137,9 +137,8 @@ This project supports PHP 5.6 as minimum version (see [Drupal 8 PHP requirements
 To prevent this you can add this code to specify the PHP version you want to use in the `config` section of `composer.json`:
 ```json
 "config": {
-    "sort-packages": true,
     "platform": {
         "php": "5.6.40"
     }
-},
```
jcnventura: "sort packages" is part of the provided composer.json. The example is correct. Please revert this.

jcnventura (changes requested):
Instead of doing a lot of unrelated changes everywhere, including some that introduce errors, please do the simple one/two line change you really want please.

And please explain what is the problem with the current setup.

README.md
phansys: This snippet is focused on the "platform" setting AFAIK, so listing other allowed configurations here seems to bring no any useful context IMHO.

composer.json
phansys: The whitelist is required in order to avoid unintended burden of the stability with dangerous dev dependencies, so a more narrow decision is to allow explicitly the non stable packages using [stability flags](https://getcomposer.org/doc/04-schema.md#package-links).

README.md
phansys: `8.0` doesn't exist really, it's the implicit SemVer [branch alias](https://getcomposer.org/doc/articles/aliases.md#branch-alias) which points to `8.x-dev`.

README.md
phansys: > Instead of doing a lot of unrelated changes everywhere, including some that introduce errors, please do the simple one/two line change you really want please.
> 
> And please explain what is the problem with the current setup.

Thanks so much for the feedback.
I don't think the changes done in this PR are unrelated, that's why they are even grouped in the same commit.
Please, let me know if you have a different perspective about this in order to try a consensus about these changes.

fafnirical: > This snippet is focused on the "platform" setting AFAIK, so listing other allowed configurations here seems to bring no any useful context IMHO.

Users would likely copy and paste this exact snippet into their `composer.json` file, which would lead to `sort-packages` being removed from their configuration.

phansys: >Users would likely copy and paste this exact snippet into their composer.json file, which would lead to sort-packages being removed from their configuration.

Based on this assumption, we should interpret [this snippet](https://github.com/drupal-composer/drupal-project/blob/ec0f41171516c18357ca4cda9068ee4cdeabaa19/README.md#should-i-commit-the-scaffolding-files) as a bad example too, didn't?
```yml
"scripts": {
    "post-install-cmd": [
        "@composer drupal:scaffold",
        "..."
    ],
    "post-update-cmd": [
        "@composer drupal:scaffold",
        "..."
    ]
},
```

fafnirical: I suppose you're right. Maybe the docs should be updated to be more clear not to remove rules like this when you're updating your config, although that should probably be a separate PR.

Maybe the README should recommend using the CLI to adjust this, rather than editing the `composer.json` file manually?

phansys: I've reverted this change, as it has nothing to do with the PR's goal. I agree that this topic should be addressed in a separate PR. Thank you so much for sharing your concerns about this.

phansys:
For those who felt confused about the motivation for these changes, I've added a more detailed example trying to explain how composer works and what my concerns are.

Thank you in advance for your feedback.

phansys:
IMHO, #512 is a good example of what happen when a package trust on an unstable dependency, and it gets even worse when that dependency is allowed blindly, like in this case.

AlexSkrypnyk:

Issue #511, opened 2019-09-02
Can't install because no matching package found


ericheydrich:
Problem 1
    - webflo/drupal-core-require-dev 8.8.x-dev requires behat/mink-selenium2-driver 1.3.x-dev -> no matching package found.
    - webflo/drupal-core-require-dev 8.7.x-dev requires behat/mink-selenium2-driver 1.3.x-dev -> no matching package found.
    - webflo/drupal-core-require-dev 8.7.6 requires behat/mink-selenium2-driver 1.3.x-dev -> no matching package found.
    - webflo/drupal-core-require-dev 8.7.5 requires behat/mink-selenium2-driver 1.3.x-dev -> no matching package found.
    - webflo/drupal-core-require-dev 8.7.4 requires behat/mink-selenium2-driver 1.3.x-dev -> no matching package found.
    - webflo/drupal-core-require-dev 8.7.3 requires behat/mink-selenium2-driver 1.3.x-dev -> no matching package found.
    - webflo/drupal-core-require-dev 8.7.2 requires behat/mink-selenium2-driver 1.3.x-dev -> no matching package found.
    - webflo/drupal-core-require-dev 8.7.1 requires behat/mink-selenium2-driver 1.3.x-dev -> no matching package found.
    - webflo/drupal-core-require-dev 8.7.0-rc1 requires behat/mink-selenium2-driver 1.3.x-dev -> no matching package found.
    - webflo/drupal-core-require-dev 8.7.0-beta2 requires behat/mink-selenium2-driver 1.3.x-dev -> no matching package found.
    - webflo/drupal-core-require-dev 8.7.0-beta1 requires behat/mink-selenium2-driver 1.3.x-dev -> no matching package found.
    - webflo/drupal-core-require-dev 8.7.0-alpha2 requires behat/mink-selenium2-driver 1.3.x-dev -> no matching package found.
    - webflo/drupal-core-require-dev 8.7.0-alpha1 requires behat/mink-selenium2-driver 1.3.x-dev -> no matching package found.
    - webflo/drupal-core-require-dev 8.7.0 requires behat/mink-selenium2-driver 1.3.x-dev -> no matching package found.
[...]

PR #512, opened 2019-09-03
Can't install because no matching package found

Unifex:
I had the semi-regular composer issue in the title and found issue #511 was current and had the solution.  It looked like there wasn't a pull request for it though.  This should resolve that.  

Credit should go to @jamie-tillman.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,6 +17,7 @@
     ],
     "require": {
         "php": ">=5.6",
+        "behat/mink-selenium2-driver": "dev-master as 1.3.x-dev",
         "composer/installers": "^1.2",
         "cweagans/composer-patches": "^1.6.5",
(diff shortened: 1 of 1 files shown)


normanlolx:
This is just a temporary fix. And this issue shouldn't be fixed here but in https://github.com/webflo/drupal-core-require-dev/pull/13.

rodrigoaguilera:
To clarify: the webflo repo only copies what is in drupal core so the problem so it should be fixed one step upstream

https://www.drupal.org/project/drupal/issues/3078671

normanlolx:
+1 for upstream. Thanks for clarifying!

Unifex:
Fair call.  If nothing has happened upstream by tomorrow I'll take a look at that.  It's the end of my day here.

PR #518, opened 2019-09-11
Deprecate PHP 7.1

MPParsley:
PHP 7.1 is no longer supported, let's remove it.

See https://www.php.net/supported-versions.php

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ dist: trusty
 sudo: false
 
 php:
-  - 7.0
   - 7.1
   - 7.2
   - 7.3
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ dist: trusty
 sudo: false
 
 php:
-  - 7.1
   - 7.2
   - 7.3
 
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         }
     ],
     "require": {
-        "php": ">=7.1",
+        "php": ">=7.2",
         "composer/installers": "^1.2",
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -3,7 +3,6 @@ dist: trusty
 sudo: false
 
 php:
-  - 7.0
   - 7.1
```
MPParsley:
```suggestion
```

composer.json
```diff
@@ -16,7 +16,7 @@
         }
     ],
     "require": {
-        "php": ">=5.6",
+        "php": ">=7.1",
```
MPParsley:
```suggestion
        "php": ">=7.2",
```

RobLoach (approved):
Likely a good call, thanks!

webflo:
Thanks for the PR!

PetersRicardo:
> PHP 7.1 is no longer supported, let's remove it.
> 
> See https://www.php.net/supported-versions.php

This is not true yet, security support is maintained till 01-12-2019 the next drupal release still falls within this time frame (and will support 7.1 as far as I can tell).

Is this intended?

goba:
This is a bit heavy handed. Since this package is the official suggested way to install Drupal with composer (https://www.drupal.org/docs/develop/using-composer/using-composer-to-install-drupal-and-manage-dependencies) it would be useful if the official version constraints would be followed. Drupal 8.8 does not even require PHP 7.1 yet let alone a minimum of 7.2.

webflo:
Thanks for your feedback. I reverted the change and we will use 7.0.8 as the new minimum.

PR #522, opened 2019-09-21
Use scaffold plugin and composer packages from Drupal core 8.8.0

greg-1-anderson:
The new drupal/core-composer-scaffold plugin in Drupal 8.8.x-dev works with earlier versions of Drupal as well. Here is an early pull request to demonstrate that.

Still to do:

- [X] Investigate installing `.editorconfig` and `.gitattributes` to the project root by default in the upstream.
- [X] ~Do not require `web-root` to be defined in the scaffold metadata; default to the parent directory of drupal/core if it is not set.~
- [x] zaporylie/composer-drupal-optimizations update https://github.com/zaporylie/composer-drupal-optimizations/pull/15/files


composer.json
```diff
@@ -19,17 +19,17 @@
         "php": ">=7.0.8",
         "composer/installers": "^1.2",
         "cweagans/composer-patches": "^1.6.5",
-        "drupal-composer/drupal-scaffold": "^2.5",
+        "drupal/core-composer-scaffold": "8.8.x-dev",
+        "drupal/core-recommended": "^8.7.0",
         "drupal/console": "^1.0.2",
-        "drupal/core": "^8.7.0",
```
zaporylie, line 24: The change above made `zaporylie/composer-drupal-optimizations` fail to determine which version of Symfony components can be removed while calculating dependencies. So either you add config manually as suggested in https://github.com/zaporylie/composer-drupal-optimizations#configuration or remove that dependency completely as it no longer makes sense to have it here.

Not sure what `drupal/core-recommended` is - new core subtree split? - but I guess I can modify `zaporylie/composer-drupal-optimizations` to also read version constraints from it. https://github.com/zaporylie/composer-drupal-optimizations/blob/master/src/Plugin.php#L23

greg-1-anderson: Thanks for noticing and commenting, @zaporylie.

Yes, what we are talking about here is a new set of subtree splits that are maintained by the drupal.org infrastructure and replace the existing community-maintained metapackages (e.g. webflo/drupal-core-strict and webflo/drupal-core-dev-dependencies).

A diagram showing the relationship between the different new metapackages can be found in the [drupal.org infrastructure package generator README](https://bitbucket.org/drupalorg-infrastructure/package-generator/src/master/README.md). It would be great if you could support projects that use these metapackages in zaporylie/composer-drupal-optimizations.

greg-1-anderson:
Defaulting the location of [web-root] will not be done; see [#3084369](https://www.drupal.org/project/drupal/issues/3084369).

greg-1-anderson:
We are working on 

- https://www.drupal.org/project/drupal/issues/3084326
- https://bitbucket.org/drupalorg-infrastructure/package-generator/pull-requests/5/project-root-assets/diff

While these are merged, we still need to rewrite the corresponding metapackage. This will probably be the last time we need to rewrite a package; we won't do it any more once we have a stable alpha release.

composer.json
zaporylie: If I read this correctly `drupal/core-recommended` is just `drupal/core` with exact version constraints for all core's required packages. Since versioning convention of the new package strictly follows `drupal/core` versions, the only thing that needs to be done is reading `drupal/core-requirements` version constraint if dependency present in project's composer.json.

https://github.com/zaporylie/composer-drupal-optimizations/pull/15/files
I'm planning to merge it later today or on Friday if no one has any objections.

greg-1-anderson:
The metapackages have been rewritten.

greg-1-anderson:
n.b: https://www.drupal.org/project/drupal/issues/3086148

The name of the scaffold metadata key in the composer.json extra section might change this week. Friday is the code freeze, so whatever is in at that point will be stable.

greg-1-anderson:
The core Drupal issue referenced above has been merged, but the legacy scaffold assets metapackage has not been rewritten yet, so Drupal 8.7.x builds using these components will fail until this is done.

greg-1-anderson:
OK, things should scaffold to the correct location again. I excluded a set of tests (Scaffold tests) that can't run here. We'll see if that gets us closer to passing; however, I also note that the tests are not passing on the master branch. I do wonder if running the Drupal tests is the best way to test this project. The Drupal test suite instructions specifically instruct that the tests should be run from a git clone + composer install, and not any other way. We may therefore have increasing failures when trying to run the same tests in the context of this project, simply because this is not a supported configuration for the tests. Also, those tests that do pass do not tell us that our project was assembled correctly, they only tell us that those tests will run in isolation, which is not necessarily relevant.

Perhaps we should just be satisfied with the site-install / runserver tests and not try to run the testsuite.

greg-1-anderson:
If you are having trouble with drupal/core-recommended, see https://www.drupal.org/project/drupal/issues/3087531

neclimdul:
I was manually testing this process to update a site. I'm having a bit of a conflict though and I'm not sure how best to resolve it.

Prior to this we used `drupal/core` which has ranged requirements on its packages. Projects are then in charge of managing their .lock file and any stricter versions of requirements. This is actually a contrib _requirement_ for `search_api_solr` as documented in its installation guide.
https://git.drupalcode.org/project/search_api_solr#installation
```
composer require symfony/event-dispatcher:"4.3.4 as 3.4.99" drupal/search_api_solr
```

`drupal/core-recommended` on the other hand locks packages at specific releases. There's a lot to be said for that I'm sure but either some additional considerations need to happen or `search_api_solr` needs to update its installation guide because you can't run that install step with this approach because `3.4.99 != 3.4.32`

greg-1-anderson:
Projects that do not wish to use pinned dependencies may use `drupal/core` instead of `drupal/core-recommended` in their composer.json file.

composer.json
```diff
@@ -19,17 +19,17 @@
...
         "drush/drush": "^9.0.0",
         "vlucas/phpdotenv": "^2.4",
         "webflo/drupal-finder": "^1.0.0",
         "webmozart/path-util": "^2.3",
         "zaporylie/composer-drupal-optimizations": "^1.0"
     },
     "require-dev": {
-        "webflo/drupal-core-require-dev": "^8.7.0"
+        "drupal/dev-dependencies": "^8.7.0"
```
webflo: drupal/dev-dependencies is marked as abandoned on packagist. should use drupal/core-dev instead?

composer.json
webflo: @zaporylie 👍 thanks for the release!

.gitignore
```diff
@@ -22,3 +22,5 @@
 
 # Ignore .env files as they are personal
 /.env
+.editorconfig
```
webflo: Hi, whats the reason to ignore those two files? Shouldn't the .editorconfig be shared with the team to ensure the same editor configuration?

composer.json
greg-1-anderson: Yeah it was renamed

.gitignore
greg-1-anderson: I think I copied that from somewhere else; I don't recall. I have no strong opinion about whether it is ignored or not. If folks were using different formatting standards they'd get a bunch of git conflicts over time, I think, though.

composer.json
```diff
@@ -19,17 +19,17 @@
         "php": ">=7.0.8",
         "composer/installers": "^1.2",
         "cweagans/composer-patches": "^1.6.5",
-        "drupal-composer/drupal-scaffold": "^2.5",
+        "drupal/core-composer-scaffold": "^8.8.0",
+        "drupal/core": "^8.8.0",
```
greg-1-anderson: The reason I originally made this ^8.7 instead of ^8.8 was to emphasize that older versions of Drupal also work. These two values will both evaluate in the same way, of course -- 8.8.0 is selected today, 8.9.0 will be selected when it is available, etc.

composer.json
```diff
@@ -19,17 +19,17 @@
...
-        "drupal/core": "^8.7.0",
         "drush/drush": "^9.0.0",
         "vlucas/phpdotenv": "^2.4",
         "webflo/drupal-finder": "^1.0.0",
         "webmozart/path-util": "^2.3",
         "zaporylie/composer-drupal-optimizations": "^1.0"
     },
     "require-dev": {
-        "webflo/drupal-core-require-dev": "^8.7.0"
+        "drupal/dev-dependencies": "^8.8.0"
```
greg-1-anderson: This should be drupal/core-dev now

webflo:
Thanks @greg-1-anderson! Updated a few other things as well.

jcnventura:
> `drupal/core-recommended` on the other hand locks packages at specific releases. There's a lot to be said for that I'm sure but either some additional considerations need to happen or `search_api_solr` needs to update its installation guide because you can't run that install step with this approach because `3.4.99 != 3.4.32`

Yes, the documentation for search_api_solr needs to be corrected to use 3.4.32 instead of 3.4.99. Otherwise, works fine.

Issue #362, opened 2018-02-26
Updating from drupal 8.4.4 to 8.4.5 do nothing due webflo/drupal-core-require-dev


calbasi:
Hi, just trying to update with:

> composer update drupal/core --with-dependencies

But composer don't update anything ;-)
Then:

> composer why-not "drupal/core:8.4.5"

Get:

> webflo/drupal-core-require-dev  8.4.4  requires  drupal/core (8.4.4)

So I should do:

> composer update drupal/core webflo/drupal-core-require-dev --with-dependencies

to update drupal core...
But I wonder if this is a "by design" behavior... I don't think so...

webflo:
True, the DX is not optimal because of this change. We should write some new documentation. 

//cc @alexpott

alexpott:
Well the good thing is that at least they are being kept in sync. I assume that the webflo/drupal-core-require-dev  requirement in composer.json was 8.4.4. Would a more relaxed requirement allow it to be updated when you update just core?

lvbeck:
same thing happens when I try to update drupal 8.4.5 to 8.5.0, composer won't update anything.

even I run composer update drupal/core webflo/drupal-core-require-dev --with-dependencies,  still no update is installed.

please provide more information about how to update drupal core from 8.4 to 8.5, thanks!

webflo:
@lvbeck https://orkjern.com/updating-to-drupal-85-with-composer should help

lvbeck:
@webflo thank you, it works after I manually edit the composer.json
```
"drupal/core": "~8.5"
"webflo/drupal-core-require-dev": "~8.5"
```
and run 
`composer update drupal/core symfony/config webflo/drupal-core-require-dev --with-dependencies`

webflo:
@lvbeck We improved the update instructions in #366 

This should help in the most cases. I keep the issue open for a couple of days, for users with trouble updating to 8.5.

rodrigoaguilera:
Maybe this really long command can be added to the "scripts" section? update-core maybe as a name?

I think is really great  the introduction of webflo/drupal-core-require-dev but I wish it had as minimal trade-offs as possible.

alexpott:
@rodrigoaguilera out of interest is there a reason you are not doing just "composer update". Most of the problems come from trying to update just one thing where actually your project has a web of dependencies.

I vote for not putting a command in scripts because you can't guarantee it is going to work. The next time it'll be something other than symfony/config

[...]

PR #524, opened 2019-09-23
Removing  "symfony/*" from update command

hansfn:
"symfony/\*" was added to the update command because of a specific problem when upgrading Drupal 8.4.4 (because of a dependency in Drupal Console or something). As discussed in #362 next time the problem can be something completely different. Keeping "symfony/\*" doesn't make any sense to me, but I might be wrong.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ new release of Drupal core.
 
 Follow the steps below to update your core files.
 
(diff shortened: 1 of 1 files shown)


phansys (approved):

webflo:
Fixed in https://github.com/drupal-composer/drupal-project/pull/522 

Thanks!

PR #525, opened 2019-09-26
Add 'contrib' to Drush commandfile installer path

greg-1-anderson:
For consistency, add 'contrib' to Drush commandfile installer path. This opens the way for folks to add site-specific commandfiles in 'custom'.

c.f. https://github.com/pantheon-systems/example-drops-8-composer/blob/2.0.0/composer.json#L89


webflo:
Thanks!

PR #534, opened 2019-12-06
Remove duplicate php requirement.

tormi:
Remove duplicate php requirement already covered in `vendor/drupal/core-composer-scaffold/composer.json`.
re: https://github.com/drupal-composer/drupal-project/commit/d2e7410c8453ff18462278836c76e56b9a973585#commitcomment-36298565

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,6 @@
         }
     ],
     "require": {
-        "php": ">=7.0.8",
         "composer/installers": "^1.2",
         "cweagans/composer-patches": "^1.6.5",
         "drupal/console": "^1.0.2",


greg-1-anderson:
Generally speaking, I prefer for a project to declare its expected minimum PHP version in its top-level composer.json file. This is a form of documentation for folks looking at the project to evaluate its requirements e.g. by browsing source on GitHub.

On a technical level, though, you are correct that it is redundant to have this, and furthermore, as a project evolves, it is possible that the top-level PHP minimum version limit might become stale over time (e.g. after upgrading to Drupal 9), so you could also argue that it would be more DRY to take this out.

petk:
+1 for keeping the root composer.json php constraint as is. This is mostly also bumped many times to even more recent version based on the code in the project.

tormi:
Fair enough.

PR #537, opened 2019-12-11
Do not use a trailing slash in web root

Sutharsan:
Athough technically the web root can have any value, the trailing slash confused me when I used drupal-project and compared it with the documentation of core-composer-scaffold. In the code implementation its value is used in the `[web-root]` placeholder. Drupal core uses the placeholder followed by a trailing slash: e.g. `[web-root]/robots.txt`. 

Some more references:
- core-composer-scaffold documentation defines it in an example as: `"web-root": "./docroot"`
- The code that performs and documents the replacement: ScaffoldFilePath::destinationPath in https://github.com/drupal/core-composer-scaffold/blob/8.8.x/ScaffoldOptions.php

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -65,7 +65,7 @@
         },
         "drupal-scaffold": {
             "locations": {
-                "web-root": "web/"
+                "web-root": "./web"
             }
         },
         "installer-paths": {


webflo:
@greg-1-anderson Please review. Thanks!

Sutharsan:
I've have created a core issue for this: [Document the correct web-root default value](https://www.drupal.org/project/drupal/issues/3101209)

PR #551, opened 2020-01-17
[#549] Update ScriptHandler.php to check for symlinks before creating files directory

### LINKED ISSUE #549: Update ScriptHandler to support symlinks
We use docker-project in our single command rollout of new projects. The /sites/default/files directory is created by the ScriptHandler and afterward the /sites/default/files directory is then symlinked by wodby for use with docker. 

This is all working fine, the problem comes with the post-update-cmd and post-install-cmd events. They are run multiple times during our creation process and createRequiredFiles check if the /sites/default/files directory exists every times. Which causes an exception because it doesn't exist it is actually a symlink now.

I would love to fix this problem and I have come up with 2 solutions:
- Check if directories are symlinks if they don't exist.
- Only run the createRequiredFiles method only in the post-create-project-cmd event instead of every time.

How do you feel about this?

SkywardRay:
Closes #549

When using drupal inside a docker container it can be useful to make the sites/default/files directory a symlink instead. For example here in the wodby documentation https://wodby.com/docs/stacks/drupal/#files_1

I propose adding a check to is_link to prevent errors caused by $fs->exists not handling symlinks.

### DIFF
--- a/scripts/composer/ScriptHandler.php
+++ b/scripts/composer/ScriptHandler.php
@@ -52,7 +52,7 @@ class ScriptHandler {
     }
 
     // Create the files directory with chmod 0777
-    if (!$fs->exists($drupalRoot . '/sites/default/files')) {
+    if (!$fs->exists($drupalRoot . '/sites/default/files') && !is_link($drupalRoot . '/sites/default/files')) {
       $oldmask = umask(0);
       $fs->mkdir($drupalRoot . '/sites/default/files', 0777);
       umask($oldmask);


normanlolx (approved):
This looks good to me. I think we should then get this in 9.x, 10.x and 11.x as well.

AlexSkrypnyk (approved):

AlexSkrypnyk:
I've pushed to `10.x` and `11.x`. I think we decided to support only current and the next version of core. Happy to discuss this offline and commit this to other branches.

Issue #558, opened 2020-04-11
Travis builds are failing due to composer-plugin-api version mismatch


shrop:
Travis builds are failing due to composer-plugin-api version mismatch. Issue is Travis is using a snapshot version of Composer that appears the have composer-plugin-api v2 and that conflicts with cweagans/composer-patches, which requires composer-plugin-api v1.

We first noticed this in same issue in Guardr, which leverages Drupal Project code. Reference: https://www.drupal.org/project/guardr/issues/3126162#comment-13549572. @joestewart resolves this by commenting out the snapshot build for Composer in the Travis config file. https://github.com/guardrdistro/guardr-project/pull/26. 

See: https://travis-ci.org/github/drupal-composer/drupal-project/jobs/673820359

```
Your requirements could not be resolved to an installable set of packages.

  Problem 1

    - cweagans/composer-patches 1.6.5 requires composer-plugin-api ^1.0 -> found composer-plugin-api[2.0.0] but it does not match your constraint.

    - cweagans/composer-patches 1.6.6 requires composer-plugin-api ^1.0 -> found composer-plugin-api[2.0.0] but it does not match your constraint.

    - cweagans/composer-patches 1.6.7 requires composer-plugin-api ^1.0 -> found composer-plugin-api[2.0.0] but it does not match your constraint.

    - Root composer.json requires cweagans/composer-patches ^1.6.5 -> satisfiable by cweagans/composer-patches[1.6.7, 1.6.6, 1.6.5].

You are using a snapshot build of Composer 2, which some of your plugins seem to be incompatible with. Make sure you update your plugins or report an issue to them to ask them to support Composer 2. To work around this you can run Composer with --ignore-platform-reqs, but this will also ignore your PHP version and may result in bigger problems down the line.

The command "composer --verbose install" failed and exited with 2 during .
```

shrop:
PR #559 should resolves this issue. All tests pass: https://travis-ci.org/github/drupal-composer/drupal-project/builds/673931674. This PR also has a fix for #561.

Nothing there could be a better fix than commenting out the composer snapshot setting. Maybe just remove that line, but not 100% on that and would love some feedback.

joestewart:
Drupal 8.x may continue to specify Composer 1.x - https://www.drupal.org/project/drupal/issues/3126566#comment-13550328

The snapshot channel seems to only be 2.x
```
 php composer.phar self-update --1 --snapshot
[...]

PR #560, opened 2020-04-12
Resolve Travis build errors with composer snapshot and php out of memory errors

shrop:
Relates to: #558


webflo:
Fixed in https://github.com/drupal-composer/drupal-project/pull/575 - Thanks!

PR #573, opened 2020-09-07
Init 9.x branch.

normanlolx:
We just need to put the travis config put.

Should we maybe set the default branch to 9.x?


normanlolx:
Ah damn the naming doesn't match anymore now.

normanlolx:
@webflo – It would definitely be easier as a maintainer. Can you maybe add me as maintainer?

webflo:
@leymannx You should have an invite.

PR #577, opened 2020-10-02
Update .env.example to reflect latest version of vlucas/phpdotenv.

mpdonadio:
The latest version of [vlucas/phpdotenv](https://github.com/vlucas/phpdotenv) doesn't support `getenv()` in `Dotenv::createImmutable()`; you need to use one of the superglobals. This updates `.env.example` to reflect proper usage.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -6,14 +6,14 @@
 # like so:
 #
 # $databases['default']['default'] = [
-#   'database' => getenv('MYSQL_DATABASE'),
+#   'database' => $_ENV['MYSQL_DATABASE'],
 #   'driver' => 'mysql',
-#   'host' => getenv('MYSQL_HOSTNAME'),
+#   'host' => $_ENV['MYSQL_HOSTNAME'],
 #   'namespace' => 'Drupal\\Core\\Database\\Driver\\mysql',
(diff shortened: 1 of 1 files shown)


weitzman:
Thats strange. This PR is a regression at least in older php versions. See https://stackoverflow.com/a/21473853/265501

mpdonadio:
If we are targeting the drupal-9.0.x branch, then the `composer.json` has a min of PHP 7.3, and drupal-drupal-8.x.y is 7.0.8, so I am not sure we need to worry too much about old PHP versions. 

From https://github.com/vlucas/phpdotenv/blob/master/README.md#putenv-and-getenv

> Using `getenv()` and `putenv()` is strongly discouraged due to the fact that these functions are not thread safe, however it is still possible to instruct PHP dotenv to use these functions. Instead of calling `Dotenv::createImmutable`, one can call `Dotenv::createUnsafeImmutable`, which will add the PutenvAdapter behind the scenes. Your environment variables will now be available using the `getenv` method, as well as the super-globals:

`load.environment.php` uses `Dotenv::createImmutable()`, so I think the doc fix is the safest.

Here is a mini example of the problem, https://github.com/mpdonadio/phpdotenv-example

PR #584, opened 2020-12-02
Allow getenv() to work again with vlucas/phpdotenv 5.x

joelpittet:
https://github.com/vlucas/phpdotenv no longer works with getenv() by default in 5.x, you need to call explicitly `Dotenv::createUnsafeImmutable` to have those calls work.  There are may places that use `getenv()` calls still including Drush.
```
-$dotenv = Dotenv::createImmutable(__DIR__);
+$dotenv = Dotenv::createUnsafeImmutable(__DIR__);
```

### DIFF
--- a/load.environment.php
+++ b/load.environment.php
@@ -11,5 +11,5 @@ use Dotenv\Exception\InvalidPathException;
 /**
  * Load any .env file. See /.env.example.
  */
-$dotenv = Dotenv::createImmutable(__DIR__);
-$dotenv->safeLoad();
+$dotenv = Dotenv::createUnsafeImmutable(__DIR__);
+$dotenv->load();
--- a/load.environment.php
+++ b/load.environment.php
@@ -10,6 +10,9 @@ use Dotenv\Exception\InvalidPathException;
 
 /**
  * Load any .env file. See /.env.example.
+ *
+ * Drupal has no official method for loading environment variables and uses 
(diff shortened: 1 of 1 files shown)


webflo:
@joelpittet Using createImmutable is the recommend ways to use library and we updated our examples in this repo accordingly. https://github.com/drupal-composer/drupal-project/commit/2d48c40ad9a8187f12fda5ee74d1830f1d0086b4

Could you try to update your project? I would review a PR to change the code in Drush.

load.environment.php
```diff
@@ -10,6 +10,9 @@
...
 /**
  * Load any .env file. See /.env.example.
+ *
+ * Drupal has no official method for loading environment variables and uses 
+ * getenv() in some places.
  */
-$dotenv = Dotenv::createImmutable(__DIR__);
-$dotenv->safeLoad();
+$dotenv = Dotenv::createUnsafeImmutable(__DIR__);
+$dotenv->load();
```
normanlolx, line 18: Is there a reason you switched `safeLoad` for `load` here? Drush site install now fails when there's no .env file around.

joelpittet: That's just the docs indicating `load()` is the way to do it. I think I need a PR for Drush too.

joelpittet: Reading the method and yeah thanks for keeping it safeLoad()

PR #550, opened 2020-01-16
drupal-composer/drupal-project has [not] been deprecated

gagarine:
As of the release of Drupal 8.8.0 - the recommend composer template changed from drupal-composer/drupal-project to the officially supported: drupal/recommended-project

I think this should be indicated in the readme.

Issue #538, opened 2019-12-16
Future of drupal-composer/drupal-project compared to drupal/recommended-project


petercossey:
Related to #535 

What's the future of this project compared to drupal/recommended-project?

If this project is going to continue, what are the points of differences and the use-cases for developers to start new projects with drupal-composer/drupal-project compared to drupal/recommended-project ?

webflo:
Hi,

i think this project has its purpose. Here are a few differences to the official template:

- Ships with drush and Drupal console
- Ships with dev dependencies (testing with phpunit)
- Drush config files for site-alias and PolicyCommands
- Configures config_sync_directory in settings.php to (config/sync)

normanlolx:
I also see great advantages in this project's [ScriptHandler](https://github.com/drupal-composer/drupal-project/blob/8.x/scripts/composer/ScriptHandler.php). Which I find an excellent starting point to include more advanced logic like for example executing other different commands to precompile assets as post install or post update commands.

Coufu:
I tried `drupal/recommended-project` and ran into issues right away when trying to install Drupal modules via `composer require drupal/foobar`. Also it didn't come with a `.gitignore`, which doesn't sound like a big deal, but it just seems like an oversight. If they didn't include a common/starter `.gitignore`, what other basic things did they not include?

Gonna stick to `drupal-composer/drupal-project` for now since it hasn't let me down. I also like that it comes with drush, drupal console, and `cweagans/composer-patches`

petercossey:
@webflo @leymannx @Coufu thanks for the info.

In summary:

drupal/recommended-project is a "bare-bones" composer template which developers can use as a starting point to build a project but which requires additional configuration for most projects.

 drupal-composer/drupal-project is a "batteries included" composer template which gives developers a Composer setup that follows current conventions for development dependencies, configuration storage and testing  (Drush, Drupal Console, .gitignore) for most Drupal projects.

Does that sound fair and accurate?

simesy:
`drupal/recommended-project` will certainly evolve (I'm sure it will get a gitignore), but it might make sense to call it "canonical".

`drupal-composer/drupal-project` is "conveniently opinionated" without being biased towards any commercial solutions, it proactively sets good practices (that I try to proactively spread by using it in projects). This project effectively proved the standards that are flowing upstream.

So maybe
`drupal/recommended-project` = core
`drupal-composer/drupal-project` = contrib

skyred:
[...]

PR #586, opened 2020-12-11
Document differences between drupal-project and drupal/recommended-project

danepowell:
I think any lay person (or even a rather experienced contributor like myself) could be mighty confused at the existence of multiple "official" project templates, namely https://github.com/drupal-composer/drupal-project and https://github.com/drupal/recommended-project

The least we can do is to acknowledge that multiple options exist, and the high level differences between them so users can make an informed choice.

Consider this an alternative to https://github.com/drupal-composer/drupal-project/pull/550

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ This project (drupal-composer/drupal-project) predates drupal/recommended-projec
 
 Drupal officially recommends using drupal/recommended-project. As it matures, drupal/recommended-project is likely to eventually replace drupal-composer/drupal-project.
 
-Whichever project template you choose, you’ll have a great experience out of the box. See [this change record](https://www.drupal.org/node/3082474) and https://github.com/drupal-composer/drupal-project/pull/550 for additional information and discussion.
+Whichever project template you choose, you’ll have a great experience out of the box. See [this change record](https://www.drupal.org/node/3082474) and [this discussion](https://github.com/drupal-composer/drupal-project/pull/550) for additional information.
 
 ### Should I commit the contrib modules I download?
 
--- a/README.md
+++ b/README.md
@@ -73,6 +73,16 @@ Follow the steps below to update your core files.
 
 ## FAQ
 
+### What’s the difference between this project and drupal/recommended-project?
+
+The two most commonly recommended Drupal project templates are this project ([drupal-composer/drupal-project](https://github.com/drupal-composer/drupal-project)) and [drupal/recommended-project](https://github.com/acquia/drupal-recommended-project).
+
+This project (drupal-composer/drupal-project) predates drupal/recommended-project and generally has more documentation, more functionality (such as Drush out of the box), and more community support.
+
+Drupal officially recommends using drupal/recommended-project. As it matures, drupal/recommended-project is likely to eventually replace drupal-composer/drupal-project.
+
+Whichever project template you choose, you’ll have a great experience out of the box. See [this change record](https://www.drupal.org/node/3082474) and https://github.com/drupal-composer/drupal-project/pull/550 for additional information and discussion.
+
 ### Should I commit the contrib modules I download?
 
 Composer recommends **no**. They provide [argumentation against but also


README.md
```diff
@@ -73,6 +73,16 @@ Follow the steps below to update your core files.
 
 ## FAQ
 
+### What’s the difference between this project and drupal/recommended-project?
+
+The two most commonly recommended Drupal project templates are this project ([drupal-composer/drupal-project](https://github.com/drupal-composer/drupal-project)) and [drupal/recommended-project](https://github.com/acquia/drupal-recommended-project).
+
+This project (drupal-composer/drupal-project) predates drupal/recommended-project and generally has more documentation, more functionality (such as Drush out of the box), and more community support.
```
webflo, line 80: Could you please rephrase this sentence and avoid using "more"? IMO, both projects are equally good. If one is better than the other this is up to reader.

How about?

> This project (drupal-composer/drupal-project) predates drupal/recommended-project and has its own documentation, own functionality (such as Drush out of the box), and community support.

Or is this too vague/confusing?

webflo:
Thanks for working on it. Really appreciated!

README.md
```diff
@@ -73,6 +73,16 @@ Follow the steps below to update your core files.
 
 ## FAQ
 
+### What’s the difference between this project and drupal/recommended-project?
+
+The two most commonly recommended Drupal project templates are this project ([drupal-composer/drupal-project](https://github.com/drupal-composer/drupal-project)) and [drupal/recommended-project](https://github.com/acquia/drupal-recommended-project).
```
webflo, line 78: The link for drupal-recommended-project is wrong. acquia => drupal

webflo (changes requested):

joachim-n:
> generally has more documentation, more functionality (such as Drush out of the box), and more community support.

This is a bit handwavy... as a user needing to decide which one to use, I'd want to know to more of the practical differences. What will using this project let me do that the other one won't? In what ways will this project restrict me that the other one won't?

README.md
danepowell: Per feedback from @webflo and @joachim-n below, do we just need to spell this out?

> The two most commonly recommended Drupal project templates are this project ([drupal-composer/drupal-project](https://github.com/drupal-composer/drupal-project)) and [drupal/recommended-project](https://github.com/drupal/recommended-project). Comparing the two:
>
> **drupal-composer/drupal-project**
> - Predates drupal/recommended-project
> - Documented via this README and [Github wiki](https://github.com/drupal-composer/drupal-project/wiki)
> - Supported via Github issues and pull requests
> - Includes additional features and dependencies such as Drush, Drupal Console, and an enforced minimum PHP version
>
> **drupal/recommended-project**
> - Officially recommended by the Drupal maintainers
> - Documented via [this change record](https://www.drupal.org/node/3082474)
> - Supported via drupal.org
>
> Both templates follow modern Composer and Drupal best practices (such as using Composer Scaffold) and provide a great out-of-the-box experience. See [this discussion](https://github.com/drupal-composer/drupal-project/pull/550) for additional information.

I understand that we don't want to editorialize, and I definitely don't want to step on any toes, but I'm not sure this dry comparison captures the fact that drupal-project seems way more user-friendly. The [Drupal recommended project](https://github.com/drupal/recommended-project) is positively spartan:
screenshot-github com-2020 12 11-15_34_06
Maybe that's because discussion happens on drupal.org, but then shouldn't the repo at least mention to go there for support?

Anyway, I'm happy as long as each project at least acknowledges the other's existence so users aren't confused.

README.md
danepowell: D'oh, yeah, I was working on acquia/drupal-recommended-project at the same time, we'd like to base it on one of these two upstream templates. I'll fix the link once we come to consensus on the wording.

README.md
ryanaslett: This should probably result in a readme in the recommended-project template directory so that it shows up on the subtree splits.

ryanaslett:
I think the primary differentiator is that drupal-project is able to be opinionated, and can ship with multiple features out of the box, whereas the drupal/recommended-project has to stick with strictly what exists, and is tested, in drupal core.  

Both projects are starting points, so theres no real limitations to either template, IMO. You get a head start on very common tools with this project. Whereas you start with a blank slate with drupal/recommended-project. 

That being said, there are sometimes trade-offs that come with the inclusion of some of these tools, for example, at one time drupal-project came with drupal/console, and when drupal 9 came out, many sites struggled to update because console was not ready for drupal 9 for many months.  Oftentimes those site owners would realize that they didnt need drupal/console in the first place, it was just one of the things that came with drupal-project, and that by removing that from their site enabled their adoption of drupal 9.    Im not really sure how to distill that kind of complexity into something easy for a lay person to digest however.

AlexSkrypnyk:
While I agree that a comparison is needed, almost every sentence in this PR needs to be updated as of 2024 to reflect the current state of community adoption of this project. We also need to make sure that there are no biased references like "most commonly" or "likely to eventually".

I'm going to close this PR for now. The requirement is still tracked in https://github.com/drupal-composer/drupal-project/issues/538 and will be addressed in a new PR.

PR #593, opened 2021-02-24
8.x branch incompatible with Composer 2.0.10.

joestewart:
Reproduce issue with Composer 2.0.10:
```
composer create-project drupal-composer/drupal-project:8.x-dev some-dir --no-interaction
cd some-dir
composer install
```


webflo:
Sorry, for the late response. Drupal 8 is EOL now. Thanks!

PR #597, opened 2021-05-07
Remove useless use statement

MatthieuScarset:


webflo:
Already fixed in 9.x. Thanks!

PR #600, opened 2021-06-07
Fixes-#602: update scripthandler file

bhanu951:
update script handler file to fix error when composer require vendor/package --no-install flag is used.

### DIFF
--- a/scripts/composer/ScriptHandler.php
+++ b/scripts/composer/ScriptHandler.php
@@ -17,46 +17,50 @@ use Webmozart\PathUtil\Path;
 class ScriptHandler {
 
   public static function createRequiredFiles(Event $event) {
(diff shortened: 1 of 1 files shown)


normanlolx:
@bhanu951 – Can you please first create an issue where you describe why someone would need to run --no-install on this template? And why you then want to have the script silently fail without any notice?

bhanu951:
Fixes #602 .

bhanu951:
@leymannx : 
> Can you please first create an issue where you describe why someone would need to run --no-install on this template? 

Done. #602 

> And why you then want to have the script silently fail without any notice?

As we are executing command with --no-install flag, no additional files will be created hence I believe there is no need of notice. If you want to add a notice we can add a notice but I believe it is not needed.

PR #628, opened 2023-10-09
For easier central logging, make sure drupal/monolog is available

jfauske:


jfauske:
This was pushed to the wrong project.

PR #633, opened 2024-01-09
Replace Drush annotations with attributes

jnoordsij:
These are available as of [Drush 12.0.0](https://github.com/drush-ops/drush/releases/tag/12.0.0). Given that a higher version is now required in the `composer.json` of this template, this should be the preferred way of referencing hooks.


normanlolx:
Oh yes, you are absolutely right, thank you!! 👍🏻

normanlolx:
For reference, code reviewed comparing to https://github.com/drush-ops/drush/blob/12.x/examples/Commands/PolicyCommands.php.

PR #648, opened 2024-05-13
[#462] Removed PHP version from `composer.json`.

### LINKED ISSUE #462: Remove php version from composer.json
Including this line in composer.json causes PHPStorm to lock certain settings like setting your PHP language level.

When I removed `"php": ">=5.6",` and ran `composer update --lock`, I could then set my language level in PHPStorm, pictured below.

screen shot 2019-02-13 at 2 04 40 pm

I think this might be a bug related to Composer/PHPStorm, but it is not a simple thing for others to find, particularly when they are configuring aspects of PHPStorm for development.

AlexSkrypnyk:
Closes #462


normanlolx (approved):
🚀

PR #649, opened 2024-05-13
[#462] Removed PHP version from `composer.json`.

### LINKED ISSUE #462: Remove php version from composer.json

AlexSkrypnyk:
Closes #462


normanlolx (approved):
Oh this looks nice also with that section in the README. 🤩

README.md
```diff
@@ -123,15 +123,36 @@ section of composer.json:
...
+`config.platform` as a _specific version_ (e.g.`8.3.1`) constraint to ensure
+that only the package versions supported by your current environment are used.
 
 ```json
 "config": {
-    "sort-packages": true,
-    "platform": {
-        "php": "8.3.1"
-    }
+  "platform": {
```
normanlolx: @AlexSkrypnyk this is indented differently than in the other MR. Wouldn't it be better we indent it the same way in both?

normanlolx (changes requested):

README.md
```diff
@@ -123,13 +123,34 @@ section of composer.json:
...
+would inherit that constraint. There is no need to add `require.php` to your
+`composer.json`.
+
+`config.platform` is a platform-specific. It is recommended to specify
+`config.platform` as a _specific version_ (e.g.`8.3.1`) constraint to ensure
+that only the package versions supported by your current environment are used.
 
 ```json
 "config": {
-    "sort-packages": true,
```
AlexSkrypnyk, line 132: removed this line as it has nothing to do with what we are trying to show here

PR #650, opened 2024-05-13
Update to use new Drupal Finder (Composer Runtime) API

jnoordsij:
Drupal Finder has been revamped to use the Composer Runtime API, which makes it much simpler in how it operates. This PR updates the composer script handler to use this new Drupal Finder API. It also adds an additional panic exit in case the root cannot be found.

See also https://github.com/webflo/drupal-finder/releases/tag/1.3.0.

*Edit:* this requires some kind of additional work for new installations, given that the `Path::makeRelative` call seems to be failing on new installations, supposedly due to `getComposerRoot` returning `null` on fresh installations. I'm not sure yet on the fix that is required here (or maybe upstream?). A repeated `composer install` run seems to fix the issues.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         "drupal/core-recommended": "^10.2.0",
         "drush/drush": "^12.4.3",
         "vlucas/phpdotenv": "^5.1",
-        "webflo/drupal-finder": "^1.2"
+        "webflo/drupal-finder": "^1.3"
     },
     "require-dev": {
         "drupal/core-dev": "^10.2.0",
--- a/scripts/composer/ScriptHandler.php
+++ b/scripts/composer/ScriptHandler.php
@@ -11,7 +11,7 @@ use Composer\Script\Event;
 use Composer\Semver\Comparator;
 use Drupal\Core\Site\Settings;
 use Drupal\Core\Site\SettingsEditor;
-use DrupalFinder\DrupalFinder;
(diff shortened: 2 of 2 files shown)


AlexSkrypnyk:
@jnoordsij 
Thank you for contributing!

Confirming that this is a real issue and that we do need to provide support for the latest version of the `DrupalFider`.

Could you please update your PR to make it pass CI.

Thank you

leymannx:
Yeah if we can get that in with the CI succeed that would be really helpful. 👌🏻

normanlolx (changes requested):
Yeah if we can get that in with the CI succeed that would be really helpful. 👌🏻

jnoordsij:
I've quickly looked into this a bit further, but I'm very much unsure what's the possible cause of the CI failure. When trying to run this locally, I'm actually presented with an error at an earlier step:
```
[Symfony\Component\Filesystem\Exception\InvalidArgumentException]
  The path "/config/sync" cannot be made relative to ".../web", because they
  have different roots ("/" and "...").
```

However, after that, repeated composer runs and/or other commands run fine.

This makes it hard for me to debug what's exactly going on; so I'll just leave this open here for now for anyone interested in navigating this furter.

normanlolx:
Fixed in https://github.com/drupal-composer/drupal-project/pull/673

PR #662, opened 2024-05-16
[#660] Fixed required Composer version constraint and added a CI job.

### LINKED ISSUE #660: Composer version requirements are not checked against Drupal version-related requirements
`ScriptHandler.php` has `checkComposerVersion()` method to check the version of Composer.

Currently, it only checks that Composer is at least version `1.0.0`. This was added in 2016 and is no longer a sufficient version for Drupal 10+. See https://www.drupal.org/docs/system-requirements/composer-requirements

>Drupal 10 will require Composer 2.3.6 or higher, since this will be the requirement for using Automatic Updates. Composer can be updated by running composer self-update.
>
>Older versions of Composer may continue to install or update Drupal successfully but bug reports will only be accepted for supported versions.

Since this check is failing to correctly check for the requirements, users with outdated version of Composer have their dependencies silently resolved to older versions, resulting in requirements conflicts:

```
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - drush/drush[12.4.3, ..., 12.x-dev] require composer-runtime-api ^2.2 -> found composer-runtime-api[2.1.0] but it does not match the constraint.
    - Root composer.json requires drush/drush ^12.4.3 -> satisfiable by drush/drush[12.4.3, ..., 12.x-dev].
```

We need to update the version constant in `checkComposerVersion()` and add a test in CI to check that the version resolution works correctly (we need a negative test).

AlexSkrypnyk:
#660

- A job was added to CI to check for negative test
__660__Fixed_required_Composer_version_constraint_and_added_a_CI_job__·_drupal-composer_drupal-project_cb50f44

- Positive tests are all our existing CI jobs that are using `stable` Composer channel

---
Once this is approved - I will create a PR for `11.x`


.github/workflows/ci.yml
```diff
@@ -85,3 +85,26 @@ jobs:
 
       - name: Run a single unit test to verify the testing setup
         run: ./vendor/bin/phpunit -c ./web/core ./web/core/modules/user/tests/src/Unit/UserAccessControlHandlerTest.php
+
+  # Using outdated Composer version to test the Composer version constraint.
+  test-composer:
```
AlexSkrypnyk, line 90: there is no simple way to use existing jobs without adding a lot of `if` to every step as this check happens early in the run.

instead, having a standalone (fast) job is easier to maintain.

---

I called this `test-composer` rather than `test-composer-fail` etc. in case we want to add more composer-related tests in the future.

But happy to change the name to anything else.

scripts/composer/ScriptHandler.php
```diff
@@ -92,8 +92,8 @@ public static function checkComposerVersion(Event $event) {
     if ($version === '@package_version@' || $version === '@package_branch_alias_version@') {
       $io->writeError('<warning>You are running a development version of Composer. If you experience problems, please update Composer to the latest stable version.</warning>');
     }
-    elseif (Comparator::lessThan($version, '1.0.0')) {
-      $io->writeError('<error>Drupal-project requires Composer version 1.0.0 or higher. Please update your Composer before continuing</error>.');
+    elseif (Comparator::lessThan($version, '2.3.6')) {
```
AlexSkrypnyk, line 95: https://www.drupal.org/docs/system-requirements/composer-requirements
Composer_requirements___System_requirements___Drupal_Wiki_guide_on_Drupal_org

normanlolx (approved):
Very nice 🤩

PR #663, opened 2024-05-17
[#278] Added support for Asset packagist.

AlexSkrypnyk:
#278

Changes:
- Added support for asset packagist.
- Added `oomphinc/composer-installers-extender` to support installation of `bower-asset` and `npm-asset` type libraries. `composer/installers` supports only `drupal-library` type.
- Added check in CI that the library is installed into the correct location.
- Added documentation to `README.md` file.

---
Once this PR is approved, I will create a similar PR for `11.x`


simesy:
It feels like asset packagist isn't universally loved.

composer.json
```diff
@@ -81,6 +89,11 @@
                 "type:drupal-drush"
             ]
         },
+        "installer-types": [
+            "bower-asset",
+            "npm-asset",
+            "drupal-library"
```
normanlolx: drupal-library is not required in that array, we could remove it from here

normanlolx (changes requested):
Nice work 🚀 only one minor thing I spotted

PR #665, opened 2024-05-17
[#278] Added support for Asset packagist.

### LINKED ISSUE #278: Add support for installing libraries via asset-packagist

AlexSkrypnyk:
Closes #278


normanlolx (approved):
Nice how composer normalize already got us covered 😊

composer.json
```diff
@@ -15,6 +15,7 @@
         "drupal/core-composer-scaffold": "^11@alpha",
         "drupal/core-recommended": "^11@alpha",
         "drush/drush": "^13@beta",
+      "oomphinc/composer-installers-extender": "^2.0",
```
normanlolx: Ah here it is this needs reindentation

normanlolx (changes requested):

Issue #266, opened 2017-04-25
Set minimum-stability to "stable"


rodrigoaguilera:
Currently minimum-stability is set to "dev".
I think we should rely on composer defaults and have the minimum stability set to "stable" just by removing that line.
https://getcomposer.org/doc/04-schema.md#minimum-stability

Then we will need to add dev, alpha, beta, RC suffixes to the specific dependencies that need it like for example devel or drupal console. 
https://getcomposer.org/doc/04-schema.md#package-links
The composer.json get also more semantic.

This way we don't hide dependencies on modules, libraries, etc that are not so mature.

This all started for me because there is a contrib module with a broken HEAD
https://www.drupal.org/node/2872222

What do you think?

webflo:
This should not happen we use ``"minimum-stability": "dev"`` **and** ``"prefer-stable": true`` for this reason. Are you sure you have both in your composer.json?

rodrigoaguilera:
I think they fixed superfish module now but I will try to publish a dummy library to reproduce this.

Anyway that problem is not the main reason for this change. Is about making dev,beta, alpha and rc dependencies more explicit instead of saying pick any stability but prefer stable.

I feel we also get a speed improvement since less releases are scanned.

slootjes:
+1 on this. It's safer and faster to specify the stability on a per package level.

rodrigoaguilera:
With drupal console stable now the template doesn't even has to specify any stability.

We can add devel as an example that is going to need a @RC tag if `"minimum-stability": "dev"` is removed

joelpittet:
Seems like core is grappling with this at the moment too:
https://www.drupal.org/project/drupal/issues/3135247

normanlolx:
Core indeed moved to [stable now (9.1.0)](https://github.com/drupal/recommended-project/blob/9.1.0/composer.json#L26). The [change record](https://www.drupal.org/node/3174472) contains a lot of good arguments to do the same. Then the README needs to be updated to explain how to properly require unstable releases.

joelpittet:
I'm on board with this too now, you can be explicit about your @dev if you need the unstable.

AlexSkrypnyk:
As per D9 change record:

[...]

PR #666, opened 2024-05-17
[#266] Set `minimum-stability` to `true` and added docs.

AlexSkrypnyk:
#266

---
Once this is approved, I will create another PR for `11.x`, but using `alpha` since D11 is not released yet.


.github/workflows/ci.yml
```diff
@@ -77,6 +77,7 @@ jobs:
       - name: Override Drupal version to dev for testing dev releases
         if: matrix.drupal-release == 'dev'
         run: |
+          composer config minimum-stability dev
```
AlexSkrypnyk, line 80: this changes the setting back to `dev` for this specific run

normanlolx (approved):
I like 🤩

PR #667, opened 2024-05-17
[#266] Set `minimum-stability` to `stable` and added docs.

### LINKED ISSUE #266: Set minimum-stability to "stable"

AlexSkrypnyk:
Closes #266


composer.json
```diff
@@ -15,6 +15,7 @@
         "drupal/core-composer-scaffold": "^11@alpha",
         "drupal/core-recommended": "^11@alpha",
         "drush/drush": "^13@beta",
+        "chi-teck/drupal-code-generator": "4.x-dev#e61b917 as 3.5.0",
```
AlexSkrypnyk: This package was updated to Symfony 7, but there is no release.

Considering that the whole `11.x` branch is "early preview", I think this should be okay to have like this.

Once Drupal 11 is in `beta` and all these packages have their releases, we can clean it up.

composer.json
```diff
@@ -10,6 +10,7 @@
         }
     ],
     "require": {
+        "chi-teck/drupal-code-generator": "4.x-dev#e61b917 as 3.5.0",
```
AlexSkrypnyk: This package was updated to Symfony 7, but there is no release.

Considering that the whole 11.x branch is "early preview", I think this should be okay to have like this.

Once Drupal 11 is in beta and all these packages have their releases, we can clean it up.

I've created an [issue](https://github.com/Chi-teck/drupal-code-generator/issues/152) - maybe we can wait for a couple of days before merging this one.

simesy: I don't think you need to add `as 3.5.0` unless another package explicitly needs that version. If you are using `composer require ...` it won't work because it checks stability. But if you just add it manually to the composer.json it works fine. This is a feature afaik.

To test, I updated it locally to be this and ran composer update. 

```
 "chi-teck/drupal-code-generator": "4.x-dev",
```

I don't know if pinning it to a commit hash is good... if the package maintainers happen to rewrite history on the dev branch then composer will stop building?

composer.json
```diff
@@ -36,7 +37,7 @@
             "url": "https://asset-packagist.org"
         }
     ],
-    "minimum-stability": "dev",
+    "minimum-stability": "alpha",
```
simesy, line 39: I like this change.

composer.json
AlexSkrypnyk: it worked for you because you probably did not use the `dev` version of core.

Drush 13.x requires `4.x-dev` for Symfony 7 compatibility https://github.com/drush-ops/drush/blob/13.x/composer.json#L36C3-L36C57

AlexSkrypnyk: `chi-teck/drupal-code-generator` is now released. waiting for Drush to start using it at `alpha+` version level.

normanlolx (approved):
Nice!! Very clear documentation, I like. I'm also in favor of waiting for new code generator release instead of getting the alias committed. But when there's a follow-up issue so we don't forget about it I'm also okay with it in the meanwhile.

composer.json
AlexSkrypnyk: https://github.com/drush-ops/drush/pull/6010

AlexSkrypnyk:
@leymannx 
We now have an issue with Drush (https://github.com/drush-ops/drush/pull/6010). I'm going to wait for a couple of days before merging the alias (ugly).

composer.json
simesy: > it worked for you because you probably did not use the dev version of core.

Sorry I isn't sinking in why this project needs to use the dev version of core.

PR #670, opened 2024-05-29
Added site information to CI.

AlexSkrypnyk:
CI was not correctly handling the installation of `dev` deps - it was installing dev deps for `10.0` not `10.2` (or actually `10.4-dev`). We were not able to pick this up because there is no information about the installed site.

In addition, currently `dev` is the "next next" version resolving `10.4-dev`, while the "next" version is `10.3@beta`. We are currently missing the "next" version.

This PR fixes all of the above.

## Changes

- Fixed `dev` version not being installed for `dev` job
- Added `beta` version to capture "next beta"
- Added installed site status information

Added_`beta`_testing_for_Drupal__·_drupal-composer_drupal-project_c68f84b


normanlolx (approved):
Thank you for figuring this out, LGTM 👏🏻

PR #673, opened 2024-06-06
Use new Drupal Finder API.

tannguyen04:


scripts/composer/ScriptHandler.php
```diff
@@ -44,7 +44,7 @@ public static function createRequiredFiles(Event $event) {
       require_once $drupalRoot . '/core/includes/install.inc';
       new Settings([]);
       $settings['settings']['config_sync_directory'] = (object) [
-        'value' => Path::makeRelative($drupalFinder->getComposerRoot() . '/config/sync', $drupalRoot),
+        'value' => '../config/sync',
```
tannguyen04, line 46: With new Drupal Finder version 1.3, $drupalFinder->getComposerRoot() return empty string not same the old one.
More info https://github.com/webflo/drupal-finder/issues/72

I think we can safety hard code relative path `../config/sync` at here. Drupal can handle relative path to Drupal root itself

AlexSkrypnyk (approved):

scripts/composer/ScriptHandler.php
normanlolx: @tannguyen04 – I was curious if this still works when you put a nested folder for `extra.drupal-scaffold.locations.web-root` (e.g. `web/test/`) but this is not how `extra.drupal-scaffold.locations.web-root` is supposed to work. You can not have nested directories at this point. So `../config/sync` should be fine.

normanlolx (approved):
LGTM 👍🏻

scripts/composer/ScriptHandler.php
normanlolx: This also fixed the tests not passing in https://github.com/drupal-composer/drupal-project/pull/650. 👌🏻

jnoordsij: Sound like waiting for a response on https://github.com/webflo/drupal-finder/issues/72 or diving into that further might be the best approach here?

sascha-krug: "safety hard code relative path ../config/sync" is breaking our install and update process because composer root, drupal root and config root are at the same level.

I will manually override the dir in our project and will wait until the webflo-issue is fixed.

normanlolx: Ahhh, of course you can have `.` as web root and then it breaks. 💡

normanlolx: @sascha-krug - would you be so kind and create a follow-up issue for it, please? I don't know when I've time at the computer again.

sascha-krug: Of course. Done: https://github.com/drupal-composer/drupal-project/issues/675

Issue #675, opened 2024-06-12
Hardcoded config dir breaks "drush updb"


sascha-krug:
After merged PR https://github.com/drupal-composer/drupal-project/pull/673  with hard coded config dir "../config/sync" our drupal install/update process breaks, because our config dir, drupal root and composer root are on the same level, meaning it should be configured as "config/sync".

For the meantime I will manually override the dir in our project.

webflo:
@sascha-krug  Could you try https://github.com/webflo/drupal-finder/pull/73 it should fix drupal-finder in composer script.

AlexSkrypnyk:
#677 has resolved this. Please feel free to re-open if the issue still exists.

PR #677, opened 2024-07-01
Changes to usage of DrupalFinderComposerRuntime API in ScriptHandler

jnoordsij:
- Add sanity null-check for `$drupalRoot`
- Restore `Path` generator for `config_sync_directory`

This is a rebased version of #650 that reverts some changes introduced in #673.
Hopefully due to the changes introduced in https://github.com/webflo/drupal-finder/pull/73 and the subsequent release (https://github.com/webflo/drupal-finder/releases/tag/1.3.1), the CI should now pass without manual overriding of the config sync directory.

This probably solves #675.

Note: if preferable, I could update `composer.json` to explicitly require `"webflo/drupal-finder": "^1.3.1"`.


scripts/composer/ScriptHandler.php
```diff
@@ -22,6 +22,15 @@ public static function createRequiredFiles(Event $event) {
     $drupalFinder = new DrupalFinderComposerRuntime();
     $drupalRoot = $drupalFinder->getDrupalRoot();
 
+    // If Drupal root was not found, exit.
```
AlexSkrypnyk: nit: follow the same code style as the rest of the file
```suggestion
    if (is_null($drupalRoot)) {
      $event->getIO()->writeError('<error>Drupal root could not be detected.</error>');
      exit(1);
    }
```

AlexSkrypnyk (changes requested):
@jnoordsij 
Thank you for your contribution!

I think this looks good. Could you please update the code to follow the code style used in the current file.

scripts/composer/ScriptHandler.php
jnoordsij: Fair point! I followed more or less the calls from `checkComposerVersion`, but this looks much cleaner!

webflo:
👍

normanlolx (approved):

normanlolx:
Glad to see this sorted out dynamically. 🤩

AlexSkrypnyk (approved):

